PR #28, opened 2017-03-21
Allow passing control channel socket via command line

avalluri:
Inbrief the change are:
- Enumerate the pollfd array for better readablility
- Add new command line option --ctrlfd to pass client control channel socket via command line
- In case client socket(-f <fd>) do not terminate the connection till the client connection is open.


src/swtpm/swtpm.c
```diff
@@ -96,6 +96,7 @@ static void usage(FILE *file, const char *prgname, const char *iface)
     "\n"
     "-p|--port <port> : use the given port\n"
     "-f|--fd <fd>     : use the given socket file descriptor\n"
+    "-F|--ctrlfd <fd> : use the given socket file descriptor for control channel\n"
     "-t|--terminate   : terminate the TPM once a connection has been lost\n"
     "-d|--daemon      : daemonize the TPM\n"
     "--ctrl type=[unixio|tcp][,path=<path>][,port=<port>[,bindaddr=address[,ifname=ifname]]][,fd=<filedescriptor]\n"
```
stefanberger: I think it would be better to pass the client file descriptor as clientfd via --ctrl.

stefanberger: Also, we would need the same change in swtpm_chardev.c.

src/swtpm/mainloop.h
```diff
@@ -54,8 +54,10 @@ struct mainLoopParams {
 #define MAIN_LOOP_FLAG_KEEP_CONNECTION (1 << 2)
 #define MAIN_LOOP_FLAG_READALL    (1 << 3)
 #define MAIN_LOOP_FLAG_END_ON_HUP (1 << 4)
+#define MAIN_LOOP_FLAG_USE_CTRLFD (1 << 5)
 
     int fd;
+    int ctrlfd;
     struct ctrlchannel *cc;
```
stefanberger: I think it would be better to hide this in struct ctrlchannel and add a function ctrlchannel_get_client_fd() to retrieve the file descriptor; pass it via --ctrl ...,clientfd=<>. Passing fd= and clientfd= should be mutually exclusive and this can be tested for in ctrlchannel.c . fd is the file descriptor of a listening socket, clientfd the file descriptor of a socket to receive data on directly. Not sure with which type of socket this will actually work.

src/swtpm/swtpm.c
```diff
@@ -231,11 +234,39 @@ int swtpm_main(int argc, char **argv, const char *prgname, const char *iface)
                         "Given file descriptor type is not supported.\n");
                 exit(1);
             }
-            mlp.flags |= MAIN_LOOP_FLAG_TERMINATE | MAIN_LOOP_FLAG_USE_FD;
+            mlp.flags |= MAIN_LOOP_FLAG_USE_FD |            /* Use this as client fd */
+                         MAIN_LOOP_FLAG_TERMINATE |         /* Terminate if requested */
+                         MAIN_LOOP_FLAG_KEEP_CONNECTION;    /* Keep connection till the client connection active */
```
stefanberger: Is this a bug you are trying to fix here?

src/swtpm/swtpm.c
```diff
@@ -231,11 +234,39 @@ int swtpm_main(int argc, char **argv, const char *prgname, const char *iface)
...
+                        "Given control channnel file descriptor type is not supported.\n");
+                exit(1);
+            }
+            mlp.flags |= MAIN_LOOP_FLAG_USE_CTRLFD;
+
+            break;
+
         case 'c':
             serverdata = optarg;
             break;
```
stefanberger: Beyond please add a description to the man page.
Most importantly, we would need a test case, something similar to test_ctrlchannel, which is already missing test cases for the 'fd' parameter. What kind of socket would it use?

src/swtpm/swtpm.c
```diff
@@ -231,11 +234,39 @@ int swtpm_main(int argc, char **argv, const char *prgname, const char *iface)
...
+             * test for wrong file types; anonymous fd's do not seem to be any of the wrong
+             * ones but are also not character devices
+             */
+            if (S_ISREG(statbuf.st_mode) || S_ISDIR(statbuf.st_mode) || S_ISBLK(statbuf.st_mode)
+                || S_ISLNK(statbuf.st_mode)) {
+                fprintf(stderr,
+                        "Given control channnel file descriptor type is not supported.\n");
+                exit(1);
+            }
+            mlp.flags |= MAIN_LOOP_FLAG_USE_CTRLFD;
```
stefanberger: This flag can then be set after handle_ctrlchannel_options() was called and 
```
if (mlp.cc != NULL && ctrlchannel_get_client_fd(mlp.cc) >=0)
    mlp.flags |= MAIN_LOOP_FLAG_USE_CTRLFD;
```

src/swtpm/swtpm.c
avalluri: My initial implementation was used '--ctrl' later i felt it would be better to match  -f , as it  is ment fir client fd :). Anyways i will do the change and re-submit.

src/swtpm/swtpm.c
avalluri: I am not sure if it was bug or feature to support clients just for a single request per connection, because i have no idea how this option been used.

But i felt with that approach we cannot support pure peer-to-peer clients, hence the change to keep the connection till the client alive,  this atleast shouldn't harm the existing clients which use -f option.

src/swtpm/swtpm.c
avalluri: I am not sure if i can provide a testcase, As tests are written using shell scripts - i have no idea how can we create and pass a unix socket to swtpm inside shell.

avalluri:
Updated the PR with below changes:
- Passing control channel clinet fd using '--ctrl clientfd='
- Updated the same in swtpm_chardev.c
- Updated the man page entry

The missing one from your comments is - test case for the change, which is bit challenging to pass clientfd inside shell script.

pohly:
@stefanberger FYI, avalluri is so kind to help me with changing the qemu patches so that they work without CUSE, by having qemu itself spawn swtpm and communicating with it via Unix domain sockets. This patch series is what's needed in swtpm itself for that to work.

The corresponding qemu patches are here: https://github.com/avalluri/qemu-tpm/tree/v280+tpm-emulator

There's currently still something not quite right, though, so let's figure that out before merging this PR here.

src/swtpm/swtpm_chardev.c
```diff
@@ -126,12 +126,17 @@ static void usage(FILE *file, const char *prgname, const char *iface)
...
     "                   the path is only valid for Unixio channels; the port must\n"
     "                   be given in case the type is TCP; the TCP socket is bound\n"
     "                   to 127.0.0.1 by default and other bind addresses can be\n"
-    "                   given with the bindaddr parameter\n"
+    "                   given with the bindaddr parameter; if fd is provided,\n"
+    "                   it will be treated as a server socket and used for \n"
+    "                   accepting client connections; if clientfd is provided,\n"
+    "                   it will be treaded as client connection;\n"
+    "                   NOTE: fd and clientfd are mutually exclusive and only\n"
+    "                   valid for UinxIO channels\n"
```
avalluri: The overall idea of '--fd, --client' fd is to use swtpm as peer-2-peer server. That allows to fork from a process and communicate using socket pairs. The intention is to start swtpm inside qemu tpm backend and communicate over unix sockets, Here are the proposed qemu changes(your review comments are most appreciated): 

 https://github.com/avalluri/qemu-tpm/commit/3d05ba019453d1659cf1d2b5aaae5db7ade5cc6f#diff-a05dbc826a94c3e144c3ab1e12d8f364R394

stefanberger: Typo: UinxIO

src/swtpm/swtpm.c
```diff
@@ -98,12 +98,17 @@ static void usage(FILE *file, const char *prgname, const char *iface)
     "-f|--fd <fd>     : use the given socket file descriptor\n"
     "-t|--terminate   : terminate the TPM once a connection has been lost\n"
     "-d|--daemon      : daemonize the TPM\n"
-    "--ctrl type=[unixio|tcp][,path=<path>][,port=<port>[,bindaddr=address[,ifname=ifname]]][,fd=<filedescriptor]\n"
+    "--ctrl type=[unixio|tcp][,path=<path>][,port=<port>[,bindaddr=address[,ifname=ifname]]][,fd=<filedescriptor>|clientfd=<filedescriptor>]\n"
```
avalluri: I updated the help text(usage()) even for swupd_chardev, As both swupd & swupd_chardev share the same parsing code, the clientfd change for control channel should effect both.

src/swtpm/ctrlchannel.c
```diff
@@ -71,7 +72,10 @@ struct ctrlchannel *ctrlchannel_new(int fd)
         return NULL;
     }
 
-    cc->fd = fd;
+    cc->fd = cc->clientfd = -1;
+    if (is_client) cc->clientfd = fd;
```
stefanberger: please format it over two lines.

pohly:
A bit related: can swtpm be configured so that it adds a prefix like "swtpm: " to each line that it outputs?

When swtpm is invoked by qemu, both share the same stderr and stdout. In such a setup it is not easy to recognize where error messages come from. For example, I just started qemu with an invalid log path for swtpm, and the output I got was:

runqemu - INFO - Running /fast/build/refkit/intel-corei7-64/tmp-glibc/work/intel_corei7_64-refkit-linux/refkit-installer-image/1.0-r0/recipe-sysroot-native/usr/bin/qemu-system-x86_64 -device virtio-net-pci,netdev=net0,mac=52:54:00:12:34:02 -netdev tap,id=net0,ifname=tap0,script=no,downscript=no -drive if=virtio,file=/fast/build/refkit/intel-corei7-64/tmp-glibc/deploy/images/intel-corei7-64/refkit-installer-image-intel-corei7-64.wic,format=raw -no-reboot -vga vmware -show-cursor -usb -usbdevice tablet -device virtio-rng-pci -drive if=pflash,format=qcow2,file=/fast/build/refkit/intel-corei7-64/tmp-glibc/deploy/images/intel-corei7-64/ovmf.qcow2 -vga std  -nographic -drive if=virtio,file=tmp-glibc/deploy/images/intel-corei7-64/my-installed-image-intel-corei7-64.wic,format=raw -tpmdev emulator,id=tpm0,tpmstatedir=tmp-glibc/deploy/images/intel-corei7-64/my-tpm,logfile=tmp-glibc/deploy/images/intel-corei7-64/my-tpm/swtpm.log,path=tmp-glibc/work/x86_64-linux/swtpm-wrappers/1.0-r0/swtpm_oe.sh -device tpm-tis,tpmdev=tpm0  -cpu Nehalem -m 256  -serial mon:stdio -serial null 
Could not open logfile for writing: No such file or directory
qemu-system-x86_64: -tpmdev emulator,id=tpm0,tpmstatedir=tmp-glibc/deploy/images/intel-corei7-64/my-tpm,logfile=tmp-glibc/deploy/images/intel-corei7-64/my-tpm/swtpm.log,path=tmp-glibc/work/x86_64-linux/swtpm-wrappers/1.0-r0/swtpm_oe.sh: tpm-emulator: probing failed : Broken pipe
qemu-system-x86_64: -tpmdev emulator,id=tpm0,tpmstatedir=tmp-glibc/deploy/images/intel-corei7-64/my-tpm,logfile=tmp-glibc/deploy/images/intel-corei7-64/my-tpm/swtpm.log,path=tmp-glibc/work/x86_64-linux/swtpm-wrappers/1.0-r0/swtpm_oe.sh: tpm-emulator: Could not cleanly shutdown the TPM: Broken pipe

The output from swtpm is the "Could not open logfile for writing: No such file or directory" in the middle.

stefanberger:
@pohly I'll look into that.

stefanberger:
@pohly Please have a look at the last 3 patches in the preview branch that add a prefix parameter to the --log option and replace all fprintf stderr output with logprintf, which now prefixes all output with 'swtpm: ' until the logging has been set up completely,

avalluri:
@stefanberger Updated the PR with required test case., stashed the change with the original commit.

src/swtpm/swtpm.c
```diff
@@ -98,12 +98,17 @@ static void usage(FILE *file, const char *prgname, const char *iface)
...
     "                   the path is only valid for Unixio channels; the port must\n"
     "                   be given in case the type is TCP; the TCP socket is bound\n"
     "                   to 127.0.0.1 by default and other bind addresses can be\n"
-    "                   given with the bindaddr parameter\n"
+    "                   given with the bindaddr parameter; if fd is provided,\n"
+    "                   it will be treated as a server socket and used for \n"
+    "                   accepting client connections; if clientfd is provided,\n"
+    "                   it will be treaded as client connection;\n"
+    "                   NOTE: fd and clientfd are mutually exclusive and only\n"
+    "                   valid for UinxIO channels\n"
```
stefanberger: Typo here: UinxIO.
besides that I think you should write '... and clientfd is only valid for UnixIO channels'

src/swtpm/swtpm_chardev.c
```diff
@@ -126,12 +126,17 @@ static void usage(FILE *file, const char *prgname, const char *iface)
...
     "                   the path is only valid for Unixio channels; the port must\n"
     "                   be given in case the type is TCP; the TCP socket is bound\n"
     "                   to 127.0.0.1 by default and other bind addresses can be\n"
-    "                   given with the bindaddr parameter\n"
+    "                   given with the bindaddr parameter; if fd is provided,\n"
+    "                   it will be treated as a server socket and used for \n"
+    "                   accepting client connections; if clientfd is provided,\n"
+    "                   it will be treaded as client connection;\n"
+    "                   NOTE: fd and clientfd are mutually exclusive and only\n"
+    "                   valid for UnixIO channels\n"
```
stefanberger: Same changes to this text here.

tests/test_clientfds.py
```diff
@@ -0,0 +1,96 @@
...
+    cmd_get_caps = bytearray([0x00,0x00,0x00,0x01])
+    expected_caps = bytearray([0x00,0x00,0x00,0x00,0x00,0x00,0x07,0xff])
+
+    def toString(arr):
+        return ' '.join('{:02x}'.format(x) for x in arr)
+
+    try:
+        ctrlfd.sendall(cmd_get_caps)
+    except SocketError as e:
+        error("SocketError")
```
stefanberger: Where does 'error' come from?

src/swtpm/swtpm.c
stefanberger: Can you please remove this change since it's not related.

pohly:
On Wed, 2017-03-29 at 13:43 -0700, Stefan Berger wrote:
> @pohly Please have a look at the last 3 patches in the preview branch
> that add a prefix parameter to the --log option and replace all
> fprintf stderr output with logprintf, which now prefixes all output
> with 'swtpm: ' until the logging has been set up completely

Thanks, looks good to me.

stefanberger:
Ok, so please address the changes in your pull request. I will then apply it to master and put my stuff into master as well and rebase the preview branch on it (with a forced push).

src/swtpm/swtpm.c
avalluri: Without this fix, in case of '-f' swtpm server is exiting just ofter serving a single request, that makes overall the 'clinentfd' patch useless :(., In my opinion this is a bug from swtpm side. Shall i put this change into a separate commit?

tests/test_clientfds.py
avalluri: some how wrong branch pushed :), i will fix this

pohly:
On Thu, 2017-03-30 at 10:12 -0700, Amarnath Valluri wrote:
> @avalluri commented on this pull request.
> 
> 
> ______________________________________________________________________
> 
> In src/swtpm/swtpm.c:
> 
> > @@ -231,11 +234,39 @@ int swtpm_main(int argc, char **argv, const char *prgname, const char *iface)
>                          "Given file descriptor type is not supported.\n");
>                  exit(1);
>              }
> -            mlp.flags |= MAIN_LOOP_FLAG_TERMINATE | MAIN_LOOP_FLAG_USE_FD;
> +            mlp.flags |= MAIN_LOOP_FLAG_USE_FD |            /* Use this as client fd */
> +                         MAIN_LOOP_FLAG_TERMINATE |         /* Terminate if requested */
> +                         MAIN_LOOP_FLAG_KEEP_CONNECTION;    /* Keep connection till the client connection active */
> 
> Without this fix, in case of '-f' swtpm server is exiting just ofter
> serving a single request, that makes overall the 'clinentfd' patch
> useless :(., In my opinion this is a bug from swtpm side. Shall i put
> this change into a separate commit?

I think the change itself is fine, but the commit message does not
explain why the change is useful and required. That needs to be added
before merging.

stefanberger:
Please put the change a separated patch. Also the comment to it should probably say 'Keep client data connection alive'.

avalluri:
Does the suggested changes:
- Keep alive client data connection change to a separate commit
- reword the SERVER/CLIENT_FD to DATA_SERVER/CLIENT_FD

stefanberger:
@avalluri @pohly Thanks! I took the patches and did some touch-ups and pushed them. So your changes are in master now. 
I added the logging prefix stuff after that and rebased the preview branch on top of the new master.

stefanberger:
@avalluri @pohly FYI: I also pushed a (forced update) to the libtpms tpm2-preview.rev142 branch. If you have permanent state files laying around from a previous version, please remove them since the layout of the permanent state has changed and the TPM 2 may go into failure mode when reading that older state.

PR #38, opened 2017-09-20
Support added to receive data socket over control socket

avalluri:
As objected by QEMU upstream developers to use two different sockets for
starting/using of swtpm, This commit adds support for passing unix domain
socket over control channel.

The summary of the changes include:
 - Defined new control command CMD_SET_DATAFD, using this clients can send data
   socket.
 - replaced read() to recvmsg() in ctrlchannel_process_fd() that should be able
   to receive ancillary data.
 - set mlp.fd and mlp.flags outside of the mainloop

Signed-off-by: Amarnath Valluri <amarnath.valluri@intel.com>


src/swtpm/swtpm_io.c
```diff
@@ -205,10 +205,9 @@ TPM_RESULT SWTPM_IO_Init(void)
     if (rc == 0) {
         port_str = getenv("TPM_PORT");
         if (port_str == NULL) {
-            logprintf(STDERR_FILENO,
-                      "SWTPM_IO_Init: Error, TPM_PORT environment variable not "
-                      "set\n");
-            rc = TPM_IOERROR;
+            TPM_DEBUG("SWTPM_IO_Init: TPM_PORT environment variable not set. "
+                      "Peer-to-peer mode enabled.\n");
+            return 0;
```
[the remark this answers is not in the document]
avalluri: ok, i will rephrase as you suggested.

src/swtpm/ctrlchannel.c
```diff
@@ -525,6 +543,31 @@ int ctrlchannel_process_fd(int fd,
         out_len = sizeof(pgc->u.resp);
         break;
 
+    case CMD_SET_DATAFD:
```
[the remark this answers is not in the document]
avalluri: ok

src/swtpm/ctrlchannel.c
```diff
@@ -56,6 +56,9 @@
 #include "tpmlib.h"
 #include "swtpm_nvfile.h"
 #include "locality.h"
+#include "mainloop.h"
+
+extern struct mainLoopParams *__mlp;
```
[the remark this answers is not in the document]
avalluri: I too thought of that, will do that change.

src/swtpm/swtpm_io.c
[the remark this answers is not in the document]
stefanberger: Also the compilation fails. See the Travis test run below.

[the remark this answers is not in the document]
avalluri: i just had a look at that, i will fix.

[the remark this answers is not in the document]
stefanberger: Another thing is that this fd transfer only works when the control socket is a UnixIO socket but not when it is a TCP socket. It's not a mistake not to give a file descriptor or path in case of UnixIO but in case of TCP it has to be given. I think you currently allow it not to be given in both cases.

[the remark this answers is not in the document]
avalluri: I think to handle this properly we should move this to parse_server_options(), so that we can exactly  know what kind of sever is configured and limit TPM_PORT environment to "tcp" server. And frankly   SWTPM_IO_ServerSocket_Open() is code duplicate of tcp_open_socket().

This leaves SWTPM_IO_Init() definition to empty. Can you please give your suggestions.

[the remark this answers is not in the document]
stefanberger: It sounds right. Can you make do this in a separate patch?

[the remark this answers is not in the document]
avalluri: sure, i am working on that.

avalluri:
@stefanberger  Regarding the testcase, i try writing it using python, but i stuck to make it working. Some how the shared file descriptor is not usable. I myself is not a pthon exprt :( Can you suggest anything, here is the code:

```
cmd_set_data_fd = bytearray([0x00,0x00,0x00,0x10])
fd, _fd = socket.socketpair(socket.AF_UNIX, socket.SOCK_DGRAM)
sock_path = os.getenv('SOCK_PATH')
ctrlfd = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
ctrlfd.connect(sock_path)
ctrlfd.sendmsg([cmd_set_data_fd],
                       [(socket.SOL_SOCKET, socket.SCM_RIGHTS,
                       bytearray(_fd.fileno()))])
fd.sendall(bytearray([0x00,0xC1,0x00,0x00,0x00,0x0E,0x00,0x00,0x00,0x15,0x00,0x00,0x00,0x0a])
)
```

stefanberger:
@avalluri Does the server get and pass the test on the shared file descriptor? What file descriptor number does the server get? Can you print that and print out the number you put in? Where does the error in the above python code occur and what is the errno code?

avalluri:
The file descriptor is received at server end with descriptor number 8. But the data written to that socket(other end of the socketpair) is not received at server side, and error/exception in client side.

stefanberger:
@avalluri What file descriptor number do you write into the UnixIO socket ? I am wondering whether the number gets rewritten on the way to fit the other process.
If you wanted to push what you have I could give it a try also.

avalluri:
I could able to figure out the issue, the issue was the way i was sending
the file descriptor, i.e., the last argument in the tuple passed to
sendmsg() should be intarray where i was sending bytearry :P.

I will update the patch with test case.

Thanks for your time.

avalluri:
@stefanberger I guess, python version on CI machine < 3.3,  that might be the testcase failure, socket.sendmsg() is only available in >= python3.3., By any chance can we update python version on CI.

tests/test_ctrlchannel
```diff
@@ -56,7 +56,7 @@ PID="$(cat $PID_FILE)"
 
 # Get the capability bits: CMD_GET_CAPABILITY = 0x00 00 00 01
 res="$(unix_tx '\x00\x00\x00\x01')"
-exp=" 00 00 00 00 00 00 07 ff"
+exp=" 00 00 00 00 00 00 17 ff"
```
stefanberger: Are you sure this is 17 and not 0f ??

tests/test_ctrlchannel
```diff
@@ -611,4 +611,27 @@ rm -f $LOG_FILE
...
+           --tpmstate dir=$TPMDIR -t --pid file=$PID_FILE \
+           --log file=$LOG_FILE,level=20 &>/dev/null &
+PID=$!
+
+if wait_for_file $PID_FILE 3; then
+	echo "Error: Socket TPM did not write pidfile."
+	exit 1
+fi
+
+LOG=$(SOCK_PATH=$SOCK_PATH exec python3 $DIR/test_setdatafd.py)
```
stefanberger: I think this test program is missing. Also, can it not be written with functions available in python 2.7?

stefanberger:
@avalluri You forgot to add the new test program to the patch...

tests/test_ctrlchannel
avalluri: For some reason, PTM_CAP_GET_CONFIG is missing in return ptm_caps, otherwise it would have been 1F FF, but now it is '17 FF'.

tests/test_ctrlchannel
avalluri: Oops, ya i just realized that i missed that file.

And regarding python2.7, the issue is there is no way to send data + sockfd in a single stretch, we needs to send in two steps:
```
   socket.sendall(data)
  multiprocessing.sendfd(socket.fileno(), fds[1].fileno)
```
This not supported by swtpm server.

tests/test_ctrlchannel
stefanberger: I'll fix that then.

stefanberger: @avalluri I fixed that now. Can you rebase?

tests/test_ctrlchannel
stefanberger: I'll try to fix this today.

avalluri: by the way off topic, have you seen the Marc's comment on my qemu patchset about passing 'locality number' in big-endian. What do you suggest

stefanberger: locality number is just a uint8_t. No big endian here.

avalluri: yes, i missed this point thanks.

stefanberger: @avalluri I have this PR now https://github.com/stefanberger/swtpm/pull/40 for receiving control command data in chunks. Please have a look and comment. If it looks good to you, let me know and I'll merge.

stefanberger: @avalluri : I created now branch 'amarnath' (https://github.com/stefanberger/swtpm/tree/amarnath) that applies (and rebases) your 2 patches from this PR on top of the 2 patches I have in PR #40. I added a patch on top of that to make things work with python 2.7. Can you clean up your 2nd patch and squash this fix-up patch?

I had to turn the order of sending the data and control messages around for python 2.7 in your python test program. This made it work. I am not sure why it works, though. In the original order it would send the control message first and swtpm's recvmsg() would return '1', which indicates that a single byte was received, but that is not possible. Subsequent waits for the data didn't work. The test case now works quite reliably with python 2.7 but it's certainly better to send the data in one chunk as python 3 seems to be capable of doing.

avalluri:
@stefanberger  How to get more info about the test failure on CI, i couldn't reproduce the failure locally.

stefanberger:
Try to apply this patch on top: https://github.com/stefanberger/swtpm/commit/e7ab994329f32be5c739bfd32e834477d0941fe3. The problem should go away then. Python 2 unfortunately is a bit racy. You can remove the time.sleep() from the python code...

stefanberger:
I kicked it off again and now it passed: https://travis-ci.org/stefanberger/swtpm/builds/281257551

stefanberger:
@avalluri I think this patch here is better for python 2:

```
iff --git a/tests/test_setdatafd.py b/tests/test_setdatafd.py
index dc9bdd4..d091451 100755
--- a/tests/test_setdatafd.py
+++ b/tests/test_setdatafd.py
@@ -5,9 +5,10 @@ import socket
 import subprocess
 import time
 import struct
+
 from array import array
 if sys.version_info[0] < 3:
-    import _multiprocessing
+    import twisted.python.sendmsg as sendmsg
 
 def toString(arr):
     return ' '.join('{:02x}'.format(x) for x in arr)
@@ -52,9 +53,10 @@ def test_SetDatafd():
         ctrlfd.connect(sock_path)
         print("Sending data fd over ctrl fd...")
         if sys.version_info[0] < 3:
-            ctrlfd.send(cmd_set_data_fd)
-            time.sleep(0.2)
-            _multiprocessing.sendfd(ctrlfd.fileno(), _fd.fileno())
+            sendmsg.sendmsg(ctrlfd, b"\x00\x00\x00\x10",
+                            [(socket.SOL_SOCKET,
+                              sendmsg.SCM_RIGHTS,
+                              struct.pack("i", _fd.fileno()))])
         else:
             ctrlfd.sendmsg([cmd_set_data_fd],
                        [(socket.SOL_SOCKET, socket.SCM_RIGHTS, fds)])
```

It adds a build-dependency on python-twisted, though; the .travis.yml would have to be adapted. I think we can merge your stuff in and then fix it.

stefanberger:
This is now merged via git am etc.

PR #40, opened 2017-09-27
swtpm: enable receiving control commands in chunks

stefanberger:
Enable receiving control commands in small chnunks. Allow 500ms
for the whole control command to be received.

Signed-off-by: Stefan Berger <stefanb@linux.vnet.ibm.com>

### DIFF
--- a/src/swtpm/ctrlchannel.c
+++ b/src/swtpm/ctrlchannel.c
@@ -253,27 +253,27 @@ struct input {
  * ctrlchannel_recv_cmd: Receive a command on the control channel
  *
  * @fd: file descriptor for control channel
- * @buffer: buffer to receive data into
- * @buffer_len: length of the buffer
+ * @msg: prepared msghdr struct for receiveing data with single
+ *       msg_iov.
  *
  * This function returns 0 or a negative number if an error receiving
  * the command occurred, including a timeout. In case of success,
  * the nunber of bytes received is returned.
  */
 static ssize_t ctrlchannel_recv_cmd(int fd,
-                                    unsigned char *buffer,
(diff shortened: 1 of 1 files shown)


src/swtpm/ctrlchannel.c
```diff
@@ -220,6 +222,157 @@ static int ctrlchannel_receive_state(ptm_setstate *pss, ssize_t n, int fd)
...
+    }
+
+    while (recvd < buffer_len) {
+        n = read(fd, &buffer[recvd], buffer_len - recvd);
+        if (n <= 0)
+            return n;
+        recvd += n;
+        /* we need to at least see the cmd */
+        if (recvd < offsetof(struct input, body))
+            goto wait_chunk;
```
avalluri: Do we really need to wait here, I mean do we need to support receiving ctrl 'command' also in chunks ?, cant we treat it as some went wrong and return.

src/swtpm/ctrlchannel.c
```diff
@@ -220,6 +222,157 @@ static int ctrlchannel_receive_state(ptm_setstate *pss, ssize_t n, int fd)
...
+
+        switch (be32toh(input->cmd)) {
```
avalluri: Just an idea: Can we take this switch out of loop, by initially try to read only sizeof(command) bytes from socket then find no of bytes still we needs to complete the request, and poll with timeout till we receive, then return.

src/swtpm/ctrlchannel.c
stefanberger: It's really only these two additional lines that allow us to even wait for that...

PR #41, opened 2017-09-29
Cleanups and avoid race in test case

stefanberger:
This patch series cleans up all python code to pass pep8 without complaints. It also removes a white space error in ctrchannel.c.

Switch the python test case that's sending the control and data message in two API calls to use the python twisted package to send it in one API call and avoid hacks on the server side.


stefanberger:
@avalluri Care to have a quick look?

avalluri:
Looks good to me.

PR #43, opened 2017-09-29
Update INSTALL file for Debian

dissoupo:


stefanberger:
Thanks. And in case `make test` runs into a problem , please pull again. I just pushed a fix to the test_ctrlchannel test case.

stefanberger:
PR #43 fixed this.

PR #45, opened 2017-10-04
tests: Some work on the tests

stefanberger:
Enable testing of 32 bit and 64 bit executables with the test suite by allowing the path to the executable for swtpm, swtpm_ioctl, and swtpm_bios to be provided via variables. Add a test script that makes use of that.

### DIFF
--- a/run_tests
+++ b/run_tests
@@ -1,0 +1,34 @@
+#!/bin/bash
+
+unset SWTPM
+
+# FIXME:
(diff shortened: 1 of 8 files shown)


stefanberger:
@avalluri Some more patches for the test suite...

avalluri:
+1 looks good to me

stefanberger:
This also uncovered some TPM 2 (libtpms) related field alignment problems for 32bit archs, which are now fixed...

PR #46, opened 2017-10-06
Convert some test cases and adapt code for Cygwin

stefanberger:
Convert and split some ctrlchannel related test cases so more tests can run on Cygwin as well.
Adapt the code related to CMD_SET_DATAFD in case of Cygwin.


stefanberger:
@avalluri Some stuff again related to your recent additions.

PR #70, opened 2018-08-03
Explicitly set python test scripts to use python2

0xabe-io:
On Linux distributions where python3 is the default python interpreter, the scripts fail to run because they have been written for python2.

### DIFF
--- a/tests/test_clientfds.py
+++ b/tests/test_clientfds.py
@@ -1,4 +1,4 @@
-#!/usr/bin/python
+#!/usr/bin/env python2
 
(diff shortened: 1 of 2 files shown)


stefanberger:
Thanks for pointing this out and for the patch. I put in the effort to make this python test code python version independent: https://github.com/stefanberger/swtpm/commit/3b31a2827f4406001837a0f6e5f02772e5b2504b

I'll pull this patch over to tpm2-preview.v2 branch shortly.

PR #108, opened 2019-01-23
Make softhsm/pkcs11 test case work on Travis on OS X

stefanberger:
We need to run the softhsm/pkcs11 test case as root (sudo) under OS X
so that we can write the file /etc/gnutls/pkcs11.conf. However, once
we run the tests as root we cannot run the 'brew ls' command anymore
since it refuses to run with high privileges. So, if we run as root we
need to use sudo to switch to the nobody user to run the 'brew ls'
command that gives us the name of the softhsm pkcs11 module.


stefanberger:
PR https://github.com/stefanberger/swtpm/pull/108 fixes the issue.

PR #113, opened 2019-02-21
swtpm-setup: fallback on $HOME/.config

elmarco:
According to the XDG spec,
https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html:
"If $XDG_CONFIG_HOME is either not set or empty, a default equal to
$HOME/.config should be used."

This fixes setting up a TPM with libvirt running in a user session.

When libvirt is running as a system instance, $HOME isn't set, so it
will fall back on @SYSCONFDIR@ (/etc usually)

Signed-off-by: Marc-André Lureau <marcandre.lureau@redhat.com>

### DIFF
--- a/samples/swtpm-localca.in
+++ b/samples/swtpm-localca.in
@@ -43,8 +43,23 @@ SETUP_TPM2_F=1
 ALLOW_SIGNING_F=2
 DECRYPTION_F=4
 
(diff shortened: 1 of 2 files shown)


stefanberger:
@elmarco It looks like you need to adapt a test case as well...

elmarco:
I missed it, because that the test isn't run on "make check" or "make distcheck". Furthermore it looks odd, as it runs with with sudo and --runas. In the end it's not clear to me which config shoudl be used. I'll propose a solution.

Issue #115, opened 2019-03-13
Linux: Seccomp profile for swtpm


stefanberger:
On Linux (only), enable a seccomp profile for preventing syscalls before entering the main processing loop. Currently the following syscalls are used by swtpm:

- exit (at shutdown)
- open (in loop, at shutdown)
- close (in loop, at shutdown)
- read (in loop)
- write (in loop)
- pipe (at setup)
- poll (in loop)
- accept (in loop)
- fstat (at setup)
- ioctl (at setup)
- signal (at setup)
- setuid (at setup)
- setgid (at setup)

Those syscalls marked with 'in loop' or 'at shutdown' need to be allowed.

The seccomp profile would be enabled by default and --with-seccomp would be the implicit default to build the project with and --without-seccomp would disable it.

Additional command line parameters could be:
```
--seccomp action=none      ; turn off seccomp profile
--seccomp action=kill      ; default behavior
--seccomp action=log       ; log the syscall 
```

`swtpm` could be started by a process that itself runs a seccomp profile and that swtpm inherits and for that reason may not be able to set its seccomp profile. To handle this case `seccomp action=none` should be used.

stefanberger:
PR #117 resolves the issue.

PR #117, opened 2019-03-14
Add seccomp profile

stefanberger:
This series of patches adds support for a seccomp profile to swtpm. It's now a blacklist approach.

This PR address issue #115.


stefanberger:
Regarding performance impace: There's a ~19% overhead on an almost pure syscall program like this one here using the extensive blacklist in this PR:

```
  int fd = open("/dev/zero", O_RDONLY);
  char data;
  read(fd, &data, sizeof(data));
  close(fd);
```

Typical programs, including swtpm, are not so heavy on syscalls, though, and therefore the impact should be less.

stefanberger:
10000 PCR extension executed from user space within a VM resulted in no performance degradation with swtpm with seccomp profile versus swtpm without seccomp profile.

PR #150, opened 2019-07-09
Report binary capability for TPM 2 data channel header

stefanberger:
Report binary capability for TPM 2 data channel header


stefanberger:
thanks. merged.

PR #151, opened 2019-07-12
swtpm: Handle EINTR for poll/read/write's

stefanberger:
Only some places for poll and read are handling the EINTR correctly
so far. Fix this now for the rest of poll/read/write's in use.

### DIFF
--- a/src/swtpm/ctrlchannel.c
+++ b/src/swtpm/ctrlchannel.c
@@ -416,9 +416,14 @@ wait_chunk:
         to = timeout.tv_sec * 1000 + timeout.tv_nsec / 1E6;
 
         /* wait for the next chunk */
-        n = poll(&pollfd, 1, to);
-        if (n <= 0)
-            return n;
+        while (true) {
+            n = poll(&pollfd, 1, to);
+            if (n < 0 && errno == EINTR)
+                continue;
+            if (n <= 0)
+                return n;
+            break;
+        }
         /* we should have data now */
     }
     return recvd;
--- a/src/swtpm/ctrlchannel.c
+++ b/src/swtpm/ctrlchannel.c
@@ -183,7 +183,7 @@ static int ctrlchannel_return_state(ptm_getstate *pgs, int fd)
                      iov[1].iov_base, min(iov[1].iov_len, 1024));
     }
 
-    n = writev(fd, iov, iovcnt);
+    n = writev_full(fd, iov, iovcnt);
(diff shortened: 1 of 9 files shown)


stefanberger:
@elmarco Any comment on this one? I think swtpm needs to handle EINTR in a loop everywhere.

elmarco:
This is not sufficient, you should handle partial read/write as well.

I suggest you introduce helpers, similar to qemu_write_full() readv_writev() etc

stefanberger:
I see QEMU's qemu_write_full() but I would have thought this function only works on file descriptors where O_NONBLOCK has been set. A test with a O_NONBLOCKing pipe shows that it's writing 64kb out of 128kb into the pipe but when NOT setting O_NONBLOCK it blocks on the write of 128kb. It works similarly with a socketpair AF_UNIX/SOCK_STREAM as we use in QEMU but here the capacity seems to be ~214kb and we get the same blocking behavior as with the pipe. We don't set O_NONBLOCK in swtpm, so we shouldn't have partial writes, but then QEMU confuses me...

Also, here from writev man page:
```
The  data  transfers  performed by readv() and writev() are atomic: the data written by writev() is written as a single block that is not intermingled with output from writes in other processes (but see pipe(7) for  an  exception);
```

We cannot have a partial write otherwise it wouldn't be atomic anymore... so it should be an all or nothing write.  But then later it says:

```
Note that it is not an error for a successful call to transfer fewer bytes than requested (see read(2) and write(2)).
```

Is this a contradiction?

elmarco:
I got that: [Linux Programming Interface](https://books.google.com/books?id=Ps2SH727eCIC&pg=PA1254&lpg=PA1254&dq=%22A+partial+read+may+occur+if+there+are+fewer+bytes%22&source=bl&ots=kMHdE6wOu4&sig=ACfU3U2OJEHDXiLnCxw0XS7iNoPMvQSZag&hl=en&sa=X&ved=2ahUKEwiZnrielLDjAhWswMQBHSgrAhUQ6AEwAHoECAEQAQ#v=onepage&q=%22A%20partial%20read%20may%20occur%20if%20there%20are%20fewer%20bytes%22&f=false
)

stefanberger:
I hope this longer series now addresses the issues.

elmarco:
ahah, you cheated with writev_full() :)

looks good,
Reviewed-by: Marc-André Lureau <marcandre.lureau@redhat.com>

PR #205, opened 2020-01-24
Enable spaces in paths

stefanberger:
Fix serveral issues in the shell scripts reported by shellcheck. We can now use file paths that have spaces in them.


elmarco:
from a quick glance, looks good to me.

PR #206, opened 2020-01-26
Some work on swtpm_cert

stefanberger:
This series of patches allows passing of passwords to swtpm_cert using environment variables, converts the parsing of command line options to use getopt_long_only and allow certificate serial numbers with larger values (up to 64bit).


elmarco:
Passing secrets via environment variable isn't ideal. We should use fd, like existing `pwdfd`

stefanberger:
> 
> 
> Passing secrets via environment variable isn't ideal. We should use fd, like existing `pwdfd`

I primarily did this to accommodate the invocation via shell scripts , but I agree, we should have this as well. Another option to try for the shell script would be to pass it via file and then use this construct here:

```
function test() { echo $@; }
test <(echo -n "test")

Output: /dev/fd/63 (NetBSD, Linux, Cygwin, Apple
Output: /tmp//sh-np.CVMkwU (named pipe): Dragonfly, FreeBSD, 
```

Besides this, I think it may be good to avoid introducing too many new command line options. So passing it via something like this here may be good to:
--signkey-password: <password>
--signkey-password: fd:<fd>
--signkey-password: file:<filename>
--signkey-password: password:<password>

elmarco:
```
--signkey-password: <password>
--signkey-password: fd:<fd>
--signkey-password: file:<filename>
--signkey-password: password:<password>
```

Looks ok, although not consistent with swtpm command line options. Although unlikely, this may also break existing users, perhaps add --signkey-pass/pwd ?

stefanberger:
@elmarco Thanks for the suggestion. Pushed an updated.

stefanberger:
Also added support for `--print-capabilities` to `swtpm_cert`.

elmarco:
looks ok to me

PR #215, opened 2020-03-01
swtpm: Add option --profile and load libtpms dynamically

joholl:
This PR is related to stefanberger/libtpms#115. To dynamically load a libtpms library (aka _profile_) during runtime, an option, `--profile name=<name>` is added to `swtpm`, `swtpm_cuse` and `swtpm_chardev`.

The profile is loaded as stated below:

- If `--profile name=<name>` is given
   - if `<name>` is `default`, load default profile `libtpms.so`/`libtpms.dylib`
   - otherwise load `libtpms-<name>.so`/`libtpms-<name>.dylib`
- If no option `--profile` was given
   - load default profile `libtpms.so`/`libtpms.dylib`

I did not implement testing, yet. What do you think?


joholl:
This PR has gone stale.

PR #283, opened 2020-07-23
swtpm: Fix link failed for swtpm and swtpm_cuse

uudiin:
swtpm and swtpm_cuse failed to compile, which is caused by not linking libtpms.so

Signed-off-by: Tianjia Zhang <tianjia.zhang@linux.alibaba.com>

### DIFF
--- a/src/swtpm/Makefile.am
+++ b/src/swtpm/Makefile.am
@@ -98,7 +98,7 @@ swtpm_CFLAGS = \
 	-DHAVE_SWTPM_CUSE_MAIN
 
 swtpm_LDADD = \
-	-L$(PWD)/.libs -lswtpm_libtpms \
+	-L$(PWD)/.libs -lswtpm_libtpms -ltpms \
 	$(LIBFUSE_LIBS) \
 	$(GLIB_LIBS) \
 	$(GTHREAD_LIBS) \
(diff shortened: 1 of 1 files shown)


src/swtpm/Makefile.am
```diff
@@ -117,7 +117,7 @@ swtpm_cuse_CFLAGS = \
 	$(HARDENING_CFLAGS)
 
 swtpm_cuse_LDADD = \
-	-L$(PWD)/.libs -lswtpm_libtpms \
+	-L$(PWD)/.libs -lswtpm_libtpms -ltpms \
```
stefanberger, line 120: Same here. Does the configure script report a warning regarding not being able to determine the crypto library used in libtpms? This would mean this it cannot find the libtpms.pc file. Also, does the following work on your system?

```
$ pkg-config --libs libtpms
-ltpms
```

If not, then libtpms got installed in the wrong directory where pkg-config does not search for .pc files. You may want to add something like `--libdir=/usr/lib64` or `--libdir=/usr/lib/x86_64-linux-gnu` to the configure line of libtpms so the libtpms.pc gets installed in the proper directory. On Fedora for example the .pc files are all in /usr/lib64/pkgconfig and therefore libtpms.pc needs to go there as well.

uudiin: Yes, `pkg-config --libs libtpms` does not find libtpms. This problem can be solved by adding `PKG_CONFIG_PATH`, I think this should make it fail in the configure phase, not in the final compilation.

stefanberger: I fixed this now in PR #286. This should solve the issue, but have a look, please.

stefanberger:
This PR is now obsolete. Closing it.

PR #286, opened 2020-07-24
build-sys: Error out if libtpms.pc cannot be found and request PKG_CO…

stefanberger:
…NFIG_PATH be set

Error out if libtpms.pc cannot be found for pkg-config. This now requires that an
in-place libtpms be accessed like this:

PKG_CONFIG_PATH=/home/stefanb/libtpms/ \
	LIBTPMS_CFLAGS=-I/home/stefanb/libtpms/include/ \
	LIBTPMS_LDFLAGS=-L/home/stefanb/libtpms/src/.libs/ \
	./configure --prefix=/usr

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -162,7 +162,12 @@ if test $? -ne 0; then
 fi
 AC_SUBST([LIBTASN1_LIBS])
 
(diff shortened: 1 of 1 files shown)


uudiin:
It works well, thanks for the patch.

PR #290, opened 2020-07-28
build-sys: fix test for netstat if tcsd is not available

diabonas:
A typo in the condition meant that `netstat` was always required regardless of whether `tcsd` is available or not.


stefanberger:
Thanks for the PR. I will merge once I know what Travis is up to...

PR #296, opened 2020-08-07
Get rid of eval when launching programs

stefanberger:
This series of patches gets rid of the usage of eval for invoking programs from within bash scripts and thus prevents potential abuse of shell features. eval is not only used for variable substitution from configuration files and the variables have to have the format '${varname}'.


elmarco:
There is too much regex and shell magic for my taste, but looks ok overall, I think

elmarco:
(fwiw, I think python scripts could be considered, if not simply a rewrite in go or rust, some day ;))

PR #317, opened 2020-09-03
Remove superflous dependency checks

diabonas:
- `expect` is no longer necessary since the Python rewrite of `swtpm_setup` in commit cc410ca91bec002eefd1c5041699b60333446714.

- The only place where `pip` is required is in [`src/swtpm_setup/Makefile.am`, l. 48](https://github.com/stefanberger/swtpm/blob/0c238a2c93cdeea4283c6e9a7b8430a9b2df2b3e/src/swtpm_setup/Makefile.am#L48), to uninstall a manually built package. Most people will have pip already installed, since it comes bundled with Python by default. It might not be installed by default in minimal build environments used for distribution packaging, but these rely on the file tracking by a package manager instead of using `make uninstall`.

    In conclusion, I think requiring pip is only helpful for a very limited number of people, and is just a superfluous build dependency for distribution packagers. Issuing a warning that the package cannot be uninstalled cleanly if pip is not found seems enough.


configure.ac
```diff
@@ -326,11 +326,6 @@ fi
 AM_CONDITIONAL([WITH_GNUTLS], [test "x$with_gnutls" = "xyes"])
 AC_SUBST([GNUTLS_LIBS])
 
-AC_PATH_PROG([EXPECT], expect)
```
stefanberger, line 329: Expect is still used here: https://github.com/stefanberger/swtpm/blob/master/samples/swtpm-create-tpmca#L61

diabonas: Good catch, though that's "just" an example (though it is indeed installed to `/usr/share/swtpm`, but not in the default path). Do you think it is worth having a hard failure in this situation?

stefanberger: It's run as a test case as well. I think this should fail hard if it's not installed.

diabonas: OK, I omitted the commit removing the check for expect and signed off the other commit.

stefanberger:
Thanks.

PR #318, opened 2020-09-03
build-sys: add configure option to disable installation of Python module

diabonas:
Distributions often have their own guidelines regarding installation of Python modules (e.g. Arch Linux [mandates using `--optimize=1`](https://wiki.archlinux.org/index.php/Python_package_guidelines#distutils) for `setup.py install`, which the Makefile in this project [does not do](https://github.com/stefanberger/swtpm/blob/0c238a2c93cdeea4283c6e9a7b8430a9b2df2b3e/src/swtpm_setup/Makefile.am#L39)). Add an option `--disable-python-installation` (disabled by default) to allow skipping the Python installation process entirely so that distributions can take care of this process manually in their preferred way.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -380,6 +380,10 @@ AS_IF([ test $? = 0 ],
        [AC_MSG_RESULT([yes])],
        [AC_MSG_ERROR([python setuptools is required])])
 
+AC_ARG_ENABLE([python-installation],
+  AS_HELP_STRING([--disable-python-installation], [Disable running setup.py install for swtpm_setup]))
+AM_CONDITIONAL([PYTHON_INSTALLATION], [test "x$enable_python_installation" != "xno"])
+
 AC_ARG_ENABLE([hardening],
   AS_HELP_STRING([--disable-hardening], [Disable hardening flags]))
 
--- a/src/swtpm_setup/Makefile.am
+++ b/src/swtpm_setup/Makefile.am
@@ -19,6 +19,7 @@ $(PY_PACKAGE): $(PY_SWTPM_SETUP_FILES)
 
 all-local: $(PY_PACKAGE)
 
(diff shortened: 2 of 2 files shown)


configure.ac
```diff
@@ -380,6 +380,10 @@ AS_IF([ test $? = 0 ],
        [AC_MSG_RESULT([yes])],
        [AC_MSG_ERROR([python setuptools is required])])
 
+AC_ARG_ENABLE([python-installation],
+  AS_HELP_STRING([--disable-python-installation], [Disable running setup.py install for swtpm_setup]))
```
stefanberger, line 384: Is this now disabled by default, as stated in the commit message, or do I have to disable it explicitly?

diabonas: The conditional `PYTHON_INSTALLATION` is true unless `$enable_python_installation` is set to `no`, so one has to explicitly use `--disable-python-installation` (or `--enable-python-installation=no`), by default the variable will be unset and therefore he conditional is false.

stefanberger:
I would merge this patch in after https://github.com/stefanberger/swtpm/pull/319. I suppose the order on the master branch doesn't matter to you. Maybe that you would like this patch applied to a stable-0.4 branch?

configure.ac
stefanberger: @diabonas Can you add your Signed-off-by to this commit, please.

diabonas: Done

diabonas:
> I would merge this patch in after #319. I suppose the order on the master branch doesn't matter to you.

Sure, I rebased the PR onto the current master.

> Maybe that you would like this patch applied to a stable-0.4 branch?

That would make things easier for me in case there is another 0.4.x release, thanks!

stefanberger:
Thanks.

PR #320, opened 2020-09-08
swtpm_setup in python part 2

stefanberger:
This PR gets rid of the C code and converts swtpm_setup into a pure python program.


elmarco (commented):
great!

PR #323, opened 2020-09-15
samples: Fix usage of passwords for private keys

stefanberger:
This PR fixes some usage aspects for passwords for the private keys of the localca. The create signing key was not properly protected with the password and when creating the platform certificate the password for the signing key was not used.


stefanberger:
This needs to be backported to stable-0.4 and stable-0.3.0

PR #346, opened 2020-10-02
swtpm: Write state files atomically using file renaming

stefanberger:
To support writing state files atomically we first write into
a temporary file and then rename it.

### DIFF
--- a/src/swtpm/swtpm_nvfile.c
+++ b/src/swtpm/swtpm_nvfile.c
@@ -132,7 +132,8 @@ static unsigned char *g_ivec;
 static TPM_RESULT SWTPM_NVRAM_GetFilenameForName(char *filename,
                                                  size_t bufsize,
                                                   uint32_t tpm_number,
-                                                 const char *name);
+                                                 const char *name,
+                                                 bool is_tempfile);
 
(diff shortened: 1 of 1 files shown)


src/swtpm/swtpm_nvfile.c
```diff
@@ -606,8 +632,13 @@ static TPM_RESULT SWTPM_NVRAM_GetFilenameForName(char *filename,        /* outpu
         break;
     }
 
-    n = snprintf(filename, bufsize, "%s/tpm%s-%02lx.%s",
-                 state_directory, suffix, (unsigned long)tpm_number, name);
+    if (is_tempfile) {
+        n = snprintf(filename, bufsize, "%s/TPM%s-%02lx.%s",
```
mz-pdm, line 636: I hope that nobody ever gets an idea to run swtpm on a case-insensitive file system...

mz-pdm (approved):
FWIW it looks good to me.

src/swtpm/swtpm_nvfile.c
stefanberger: Will change this to `TMP` :-)

mz-pdm (approved):

PR #368, opened 2020-11-11
build-sys: fix configure python cryptography error message

Etsukata:
Currently it shows "python setuptools is required" if "import cryptography" failed.


stefanberger:
Thanks!

Issue #376, opened 2020-11-24
Partial read support for swtpm-cuse


AndreasFuchsSIT:
This is half a Question, half a bug and half a feature reuest.

Does swtpm-cuse support partial reads of responses as the tpm kernel driver does ?

I think we might have run into this issue here: https://github.com/tpm2-software/tpm2-tss/issues/1918
It looks like the TPM is sending the header twice on two subseuent reads, instead of resuming after the response header.

Some background; with 3.0 we switched to partial read support by default / autodetection.

As a fix on our side: is there a way to distinguish the swtpm-cuse from a real tpm so we could add this to the autodetection code ?

Could you add partial-read support to swtpm ?

stefanberger:
My suggestion would be to use `swtpm chardev --vtpm-proxy` and you get the kernel driver's partial read support immediately.

AndreasFuchsSIT:
Noted, that makes a lot of sense.
Is there still a way for tpm2-tss to detect swtpm-cuse so we handle legacy-users of swtpm-cuse more gracefully ?

stefanberger:
> Is there still a way for tpm2-tss to detect swtpm-cuse so we handle legacy-users of swtpm-cuse more gracefully ?

To avoid partial reads in case swtpm-cuse is being used? It offers the control channel ioctl's on the `/dev/tpm%d` interface, and won't have a `/dev/tpmrm%d`. The former would probably be easier to use for probing, but then maybe it's just a matter of documenting it to steer users away from swptm-cuse.

stefanberger:
I now created PR #394 for support of partial reads. Can you test it?

stefanberger:
I merged the extension. It comes with a test case and should be good to go.

PR #394, opened 2020-12-29
swtpm_cuse: Support parial reads (issue #376)

stefanberger:
This patch adds support for partial reads to the CUSE swtpm.

We introduce a ptm_read_offset variable that holds the offset where to read
from next. It is reset every time a command has been processed as part of a
write() so that subsequent read()s start reading from offset 0. It is
advanced by the number of bytes that were read.

### DIFF
--- a/src/swtpm/cuse_tpm.c
+++ b/src/swtpm/cuse_tpm.c
@@ -91,6 +91,9 @@ static unsigned char *ptm_request;
 /* buffer containing the TPM response */
 static unsigned char *ptm_response;
 
+/* offset from where to read from; reset when ptm_response is set */
+static size_t ptm_read_offset;
+
 /* the sizes of the data in the buffers */
 static uint32_t ptm_req_len, ptm_res_len, ptm_res_tot;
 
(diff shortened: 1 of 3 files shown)


stefanberger:
Some primitive tests with sending `TPM2_Startup` and reading partial results (several times) show that it's working:

```
$ sudo swtpm cuse --tpmstate dir=. --tpm2 -n vtpm0
$ swtpm_ioctl -i /dev/vtpm0
$ sudo chmod 777 /dev/vtpm0
$ exec 100<>/dev/vtpm0
$ echo -en '\x80\x01\x00\x00\x00\x0c\x00\x00\x01\x44\x00\x00' >& 100
$ dd bs=1 count=4 <&100 2>/dev/null | od -tx1
0000000 80 01 00 00
0000004
$ dd bs=1 count=4 <&100 2>/dev/null | od -tx1
0000000 00 0a 00 00
0000004
$ dd bs=1 count=4 <&100 2>/dev/null | od -tx1
0000000 01 01
0000002
$ echo -en '\x80\x01\x00\x00\x00\x0c\x00\x00\x01\x44\x00\x00' >& 100
$ dd bs=1 count=4 <&100 2>/dev/null | od -tx1
0000000 80 01 00 00
0000004
$ echo -en '\x80\x01\x00\x00\x00\x0c\x00\x00\x01\x44\x00\x00' >& 100
$ dd bs=1 count=4 <&100 2>/dev/null | od -tx1
0000000 80 01 00 00
0000004
$ dd bs=1 count=4 <&100 2>/dev/null | od -tx1
0000000 00 0a 00 00
0000004
$ dd bs=1 count=4 <&100 2>/dev/null | od -tx1
0000000 01 01
0000002
```

Issue #415, opened 2021-02-19
swtpm_ioctl can hang


elmarco:
Since commit abadb408a92e23dd2f9088909d7a3f183bf8a847, swtpm no longer listen for extra clients.

This has several consequences, such as swtpm_ioctl hanging, notably when libvirt tries to qemuTPMEmulatorStop() and hangs.

See also https://bugzilla.redhat.com/show_bug.cgi?id=1914643

elmarco:
Maybe qemuTPMEmulatorStop() shouldn't be done by libvirt, but by qemu instead since it holds the control channel.

stefanberger:
I am starting VMs all the time and did not notice any change. Strange. How did you notice?

elmarco:
Just try saving the VM, it will hang libvirt.

stefanberger:
I have done that as well, just yesterday.

stefanberger:
What commands are you using to save a VM? I am doing it from virt-manager level (arrow-down button + menu item save). It works just fine for me.

elmarco:
virsh managedsave vm, or from the virt-manager UI, "save" the vm.

stefanberger:
I cannot reproduce this problem:
```
# virsh dumpxml 2 | grep tpm
    <tpm model='tpm-crb'>
      <alias name='tpm0'/>
    </tpm>
# virsh managedsave 2

Domain 2 state saved by libvirt
```

Libvirt blocks until the state is saved.

elmarco:
hmm, that might be a regresison with libvirt then, not waiting for qemu to finish before doing TPM stop.

Both machines I test on have libvirt/virsh 7.0

stefanberger:
There are two calls to qemuTPMEmulatorStop():

- https://gitlab.com/libvirt/libvirt/-/blob/master/src/qemu/qemu_tpm.c#L770   A cleanup of a left-over before starting a new one.
- https://gitlab.com/libvirt/libvirt/-/blob/master/src/qemu/qemu_tpm.c#L859   As part of teardown of external devices that from the perspective of swtpm has to be called after QEMU has terminated. In the source code I am looking at this seems to be the case (libvirt 6.1).

elmarco:
The order of qemu kill vs TPM stop got changed in:

commit 5cde9dee8c70b17c458d031ab6cf71dce476eea2
Author: Masayoshi Mizuma <m.mizuma@jp.fujitsu.com>
[...]

PR #417, opened 2021-02-20
swtpm_ioctl: Use timeouts for communicating with swtpm (Unix socket)

stefanberger:
Start using timeouts when communicating with swtpm over Unix sockets so
that swtpm_ioctl does not just hang when it cannot communicate with swtpm
such as establishing a connection or reading results. This is because swtpm
listens to nlye one control channel connections.

This patch addresses an aspect of the problem reported in issue #415
but may also starting hiding bugs if certain operations are done in
the wrong order, as was the case in libvirt (6.10 & 7.0).

### DIFF
--- a/src/swtpm_ioctl/tpm_ioctl.c
+++ b/src/swtpm_ioctl/tpm_ioctl.c
@@ -67,6 +67,7 @@
 #include <netdb.h>
 #include <sys/param.h>
 #include <signal.h>
+#include <poll.h>
 
 #include <swtpm/tpm_ioctl.h>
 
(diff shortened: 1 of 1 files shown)


stefanberger:
@elmarco Let me know if you have any comments. I do not intend to back port it. I think the fact that it was blocking actually helped detect the libvirt issue.

elmarco:
@stefanberger I agree with you hanging makes debugging easier. At the same time, in production for such auxiliary VM task, libvirt shouldn't become unresponsive.

Perhaps this patch isn't helping then, and we should look at improving libvirt code instead (using a thread for subprocess communcation?).

stefanberger:
I merged it now anyway. Libvirt had a regression that needed to be fixed anyway.

Issue #420, opened 2021-02-23
Running 'make check' fails due to Error creating local CA's signing key and cert.


lihanshang:
Hey 

**Describe the bug**
I am running the tests in my docker container and see this error. I found the exact same error logs in this issue: https://github.com/stefanberger/swtpm/issues/382

```
FAIL: test_swtpm_setup_create_cert
==================================

TPM is listening on Unix socket.
Error: Could not run /tmp/swtpm/src/swtpm_setup/swtpm_setup.
Setup Logfile:
Starting vTPM manufacturing as root:root @ Mon 22 Feb 2021 11:51:30 PM GMT
Successfully created EK.
  Invoking /tmp/swtpm/samples/swtpm-localca --type ek --ek bfabb35393f872406fd0238bef9d7d14814aaac8a08c849996cc38954cf0aa46ee5aa509b2e279b50e765337f8061e9a53d22a96bea6785ae579d5bac5be681d1a496e008d03722e799f212d3584d37cc29f60ae2815bc6298f2d72630c21c567cbe85941be2e3ad842c4d987f0c159357ae0f152451b8051a56c9c123a4a9fc931a5f876ec5ce31afa7cd252db776a2e6d8475138e3f15a02c703a9d3532ba9d81ca17e8a57bef35ed1e7439246ae8c405f5958ede181a3a006b83f0ec213d6070e0f746aca0f864f56b55d20877e75d25e277eb46a36e2917c7e110bff323c564517e62bc3897f5b80b7fad111de12682ef4452ef3ffb5b0424e944b90e8ab --dir /tmp/tmp.UJaGb96mS7 --logfile /tmp/tmp.UJaGb96mS7/logfile --tpm-spec-family 1.2 --tpm-spec-level 2 --tpm-spec-revision 116 --tpm-manufacturer id:00001014 --tpm-model swtpm --tpm-version id:00740001 --configfile /tmp/tmp.UJaGb96mS7/swtpm-localca.conf --optsfile /tmp/tmp.UJaGb96mS7/swtpm-localca.options
Creating root CA and a local CA's signing key and issuer cert.
Could not create root-CA
Generating a self signed certificate...
importing --load-privkey: /tmp/tmp.UJaGb96mS7/swtpm-localca-rootca-privkey.pem: Decryption has failed.
Error creating local CA's signing key and cert.
Error returned from command
An error occurred. Authoring the TPM state failed.
[...]

PR #428, opened 2021-03-15
build-sys: Check for minimum required gnutls 3.4.0

stefanberger:
RHEL 7's gnutls 3.3.29 does not take the private key passwords like later
versions take it. We require at least 3.4.0, though I am not entirely sure
when that change occurred. We may actually require >3.4.0.


stefanberger:
This change addresses the issue in #420 directing users to a later version of gnutls.

PR #451, opened 2021-06-03
Clean up after the rewrite from Python to C

andreabolognani:


andreabolognani:
Never mind, I just noticed that https://github.com/stefanberger/swtpm/pull/450 already contains most of these changes 🤦

stefanberger:
Yes, gcc-11 got picky but not necessarily smarter... Trying scratch builds now. Thanks anyway.

PR #452, opened 2021-06-03
Improvements to rpm build

andreabolognani:
Plus a couple of extras :)

While working on #451, I struggled to validate my changes: after this PR, the process should hopefully result a bit more straightforward.


stefanberger (commented):
The reason why I was keeping swtpm.spec around was to be able to install dependencies via the spec file without having to go through configure.

stefanberger:
Thanks.

andreabolognani:
> The reason why I was keeping swtpm.spec around was to be able to install dependencies via the spec file without having to go through configure.

Oh yeah, that makes sense... Now that you mention it, I forgot to update the `INSTALL` file to match my changes!

Do you want me to reintroduce the `swtpm.spec` file? Alternative approaches that I can think of:

* document a shell command that grabs all `BuildRequires` from `swtpm.spec.in` and prints them out in a form that can be passed directly to `yum install`;
* create an `INSTALL.RPM-DEPS` file that contains the list of dependencies; we could also have a matching `INSTALL.DEB-DEPS` file and point users to those, making the list of packages as it currently exists in `INSTALL` (as well as the note about `build-essential` etc.) unnecessary.

Let me know which approach you prefer and I'll try to get it done :)

stefanberger:
I would avoid another file with the same content as the spec file. I can also re-introduce swtpm.spec now, not a problem. I you need to update `INSTALL`, please send a PR.

PR #460, opened 2021-06-15
swtpm: Fix SWTPM names in comments and messages

franciozzy:
Some of the code comes from libtpms, where various methods are named
"TPM_Something". The swtpm version of these methods are named
"SWTPM_Something". However, certain debug/log messages and comments were
updated accordingly to reflect that.

This is a cosmetic change that fixes that.

Signed-off-by: Felipe Franciosi <felipe@nutanix.com>


stefanberger:
Thanks.

PR #470, opened 2021-06-23
swtpm: Issue fsync to ensure state data reaches disk

Etsukata:
Add fsync(2) before rename(2) for temp file to ensure data reaches disk
and for directory which containing state file to ensure directory entry
also reaches disk.

### DIFF
--- a/src/swtpm/swtpm_nvfile.c
+++ b/src/swtpm/swtpm_nvfile.c
@@ -457,6 +457,7 @@ SWTPM_NVRAM_StoreData_Intern(const unsigned char *data,
     uint32_t      lrc;
     int           irc;
     int           fd = -1;
+    int           dir_fd = -1;
     char          tmpfile[FILENAME_MAX];  /* rooted temporary file */
     char          filename[FILENAME_MAX]; /* rooted file name from name */
     unsigned char *filedata = NULL;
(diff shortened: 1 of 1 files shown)


Etsukata:
Ref: LWN "Ensuring data reaches disk" https://lwn.net/Articles/457667/

src/swtpm/swtpm_nvfile.c
```diff
@@ -532,6 +533,17 @@ SWTPM_NVRAM_StoreData_Intern(const unsigned char *data,
             rc = TPM_FAIL;
         }
     }
+    if (rc == 0 && fd >= 0) {
+        TPM_DEBUG("  SWTPM_NVRAM_StoreData: Syncing file %s\n", tmpfile);
+        irc = fsync(fd);
```
stefanberger, line 538: I understand this part.

src/swtpm/swtpm_nvfile.c
```diff
@@ -557,6 +569,39 @@ SWTPM_NVRAM_StoreData_Intern(const unsigned char *data,
         }
     }
 
+    if (rc == 0 && fd >= 0) {
+        TPM_DEBUG(" SWTPM_NVRAM_StoreData: Opening dir %s\n", state_directory);
+        dir_fd = open(state_directory, O_RDONLY);
```
stefanberger, line 574: But why do we need to sync on the whole directory if we are only writing a file? Can you give me a pointer to read why this would be necessary?

Etsukata: Sure.
Man fsync(2): https://man7.org/linux/man-pages/man2/fdatasync.2.html

> Calling fsync() does not necessarily ensure that the entry in the
> directory containing the file has also reached disk.  For that an
> explicit fsync() on a file descriptor for the directory is also
> needed.

stefanberger: Can you add a comment to the code that indicates that this is necessary, please.

Etsukata: Sure, done.

src/swtpm/swtpm_nvfile.c
```diff
@@ -532,6 +533,17 @@ SWTPM_NVRAM_StoreData_Intern(const unsigned char *data,
             rc = TPM_FAIL;
         }
     }
+    if (rc == 0 && fd >= 0) {
+        TPM_DEBUG("  SWTPM_NVRAM_StoreData: Syncing file %s\n", tmpfile);
+        irc = fsync(fd);
+        if (irc != 0) {
+            logprintf(STDERR_FILENO,
+                      "SWTPM_NVRAM_StoreData: Error (fatal) syncing file\n");
```
pksnx, line 541: Would be good to print the _errno_ value for this and the other _fsync()_ calls below.

Etsukata: Reasonable. I'll change that.

stefanberger:
Thanks.

Etsukata:
This PR caused AppArmore/libvirt issue https://github.com/stefanberger/swtpm/issues/484

Issue #461, opened 2021-06-16
Feature proposal: pluggable storage backends


franciozzy:
**Overview**
Today SWTPM only supports writing TPM state as files in a filesystem. That is too restrictive for deployments that store VM data in external storage infrastructures accessible via NVMeOF, iSCSI, S3, &c. Using external storage is particularly relevant for VMs that are protected by high-availability features, where a host crash is handled by VMs restarting elsewhere. In those cases, the TPM data *must* be synchronously stored outside of the host.

Mounting external storage directly on the host and using it with a general purpose filesystem (so that SWTPM can be used AS IS) can violate security certifications and has other operability, scalability, and maintainability concerns. It is much preferable to plug a storage driver directly in SWTPM and have it accessing the external storage directly from userspace.

**Proposal**
We would like to propose an abstraction layer with a similar (or identical) interface as offered by [swtpm_nvfile.h](https://github.com/stefanberger/swtpm/blob/master/src/swtpm/swtpm_nvfile.h). The default implementation would continue to work as is, but plugins could be written to implement arbitrary storage backends. One possibility is to enhance the SWTPM command line interface to indicate which storage plugin to use.

We would like to use this Issue to discuss the architecture specification / design of this feature request. If the community is happy to see some code, we will gladly publish an RFC as a PR or separate GH branch on a fork.

stefanberger:
Can you rename the title to 'Feature proposal'?

I understand the support for S3. Presumably one formulates a per-VM URL where to store the state under. How is this supposed to work with iSCSI and multiple VMs' vTPM state and assignments and management of blocks/sectors on a storage device? I know that you are very interest in this type of backend, so I'd be curious to learn how this is supposed to work.

Otherwise my suggestion would be that we define an interface for the storage and push all the storage behind that interface, leaving the encryption/decryption to the core code in swtpm and all data before it reaches the backends for writing is encrypted if the user chooses so. We could compile the storage backends as shared libraries and load them as needed.

[...]

PR #490, opened 2021-07-20
swtpm: Abstract NVRAM interface for pluggable state store

Etsukata:
Related GH issue: https://github.com/stefanberger/swtpm/issues/461

This is RFC PR to add an abstraction layer to NVRAM state store
implementation for pluggable storage backends. No functional changes are 
intended in this change. The default state store backend ("file backend")
keeps current behavior.

To make swtpm ready for pluggable store, this patch moves file related
operations to the separate file (swtpm_nvstore_file.c) and defined the 
interface for plugins (nvram_backend_ops in swtpm_nvstore.h). The 
interface can be used by each plugin which will be added later.

With the interface, each plugin can access its "backend_uri" which
points to the location of the backend storage, for example S3 bucket
URL or iSCSI URL, and decide how it stores TPM state data.

The second commit adds two options to specify pluggable backend type and
backend URI.

Ex:
  --tpmstate backend=...,backend_uri=...

Backend URI is specific to each backend plugin which points to the 
location of the NVRAM files.
Currently, "file" is the only one available backend. In this case
backend_uri should be a path to the directory where files are stored.

This option is designed to compatible with existing "dir" option.
If "dir" is specified, swtpm prioritize "dir" ignoring "backend".


stefanberger:
It definitely looks manageable.
What we would also need is entries in man pages and an extension to

Etsukata:
Thanks for the comment. I'll add entries in man pages and an extension to `--print-capability`.

tests/test_commandline
```diff
@@ -174,4 +174,66 @@ fi
...
+$SWTPM_EXE socket \
+	-p $PORT \
+	--tpmstate backend=file,backend_uri=$TPMDIR,mode=$FILEMODE \
+	--pid file=$PID_FILE \
+	--flags not-need-init \
+	${SWTPM_TEST_SECCOMP_OPT} &
+PID=$!
+
+if wait_port_open $PORT $PID 4; then
+	echo "Test 1 failed: TPM did not open port $PORT"
```
stefanberger, line 191: 'Test 4' ... here and below

src/swtpm/common.c
```diff
@@ -620,16 +633,30 @@ parse_tpmstate_options(char *options, char **tpmstatedir, mode_t *mode)
...
+            goto error;
+        }
+        *tpmstatebackend_uri = strdup(backend_uri);
+        if (!*tpmstatebackend_uri) {
+            logprintf(STDERR_FILENO, "Out of memory.");
+            goto error;
+        }
+    } else {
+        logprintf(STDERR_FILENO,
+                  "The file parameter is required for the tpmstate option.\n");
```
stefanberger, line 659: it's not the file parameter but the directory or backend+backenduri parameters

src/swtpm/common.c
```diff
@@ -620,16 +633,30 @@ parse_tpmstate_options(char *options, char **tpmstatedir, mode_t *mode)
...
+        *tpmstatebackend = strdup(backend);
+        if (!*tpmstatebackend) {
+            logprintf(STDERR_FILENO, "Out of memory.");
+            goto error;
+        }
+        *tpmstatebackend_uri = strdup(backend_uri);
+        if (!*tpmstatebackend_uri) {
+            logprintf(STDERR_FILENO, "Out of memory.");
+            goto error;
+        }
```
stefanberger, line 656: error label should free *tpmstatebackend.
You could also move the two strdup's together and then do
`if (!*tpmstatebackend || !*tpmstatebackenduri) { ...`
to make this a bit shorter,

stefanberger:
Can you please also rebase this on the latest master?

tests/test_commandline
Etsukata: Oops, thanks.

src/swtpm/common.c
Etsukata: Thanks, I'll fix the error message.

src/swtpm/common.c
Etsukata: Thanks, let me use your shorter way.
`*tpmstatebackend` is also freed in caller (`handle_tpmstate_options()`) but it may be clearer to add free in error label too. I'll add them.

Etsukata:
> Can you please also rebase this on the latest master?

Sure, rebased.

tests/test_commandline
```diff
@@ -174,4 +174,66 @@ fi
 
 echo "Test 3 passed"
 
+# Test 4: --tpmstate backend=file parameters test
```
stefanberger, line 177: I think you'll need to add a call to `cleanup` before the `echo` above to clean up the TPMDIR created in line 107.

tests/test_commandline
```diff
@@ -174,4 +174,66 @@ fi
...
+if wait_for_file $PID_FILE 3; then
+	echo "Error: ${SWTPM_INTERFACE} TPM did not write pidfile."
+	exit 1
+fi
+
+validate_pidfile $PID $PID_FILE
+
+${SWTPM_BIOS} &>/dev/null
+if [ $? -ne 0 ]; then
+	echo "Test 4 failed: tpm_bios did not work"
```
stefanberger, line 210: tpm_bios -> swtpm_bios

stefanberger: or rather `${SWTPM_BIOS}`

tests/test_commandline
Etsukata: Correct. I'll add `cleanup` just after `echo "Test 3 passed"` as Test 1 does.

tests/test_commandline
Etsukata: Thanks, I'll fix that and also "Test 1" as well.

man/man8/swtpm.pod
```diff
@@ -317,6 +323,10 @@ The I<--key> option supports the I<fd=> parameter.
 
 The I<--key> option supports the I<pwdfd=> parameter.
 
+=item B<nvram-backend-file>
+
+The I<--tpmstate> option supports the I<backend=file> option.
```
stefanberger, line 328: ... supports the I<backend=file> and I<backend_uri=uri> parameters.

I hope that everything that's needed by the backends can be encoded into the backend_uri parameter...

src/swtpm/swtpm_nvstore_file.c
```diff
@@ -0,0 +1,468 @@
...
+    .delete  = SWTPM_NVRAM_DeleteName_File,
+};
+
+/* SWTPM_NVRAM_GetFilenameForName() constructs a rooted file name from the name.
+
+   The filename is of the form:
+
+   state_directory/tpm_number.name
+
+   A temporary filename used to write to may be created. It shold be rename()'d to
```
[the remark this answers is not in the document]
Etsukata: Thanks

man/man8/swtpm.pod
Etsukata: Yes, we can encode parameters in uri for example `backend_uri=iscsi://...,key1=val1,key2=val2`.

Etsukata:
> We'll need similar extensions for the swtpm_setup tool as well to choose the backend ...

Yes, we need similar extensions in swtpm_setup as well. Implementing the extension as it is (i.e. adding abstraction layer to swtpm_setup) may not be a good idea because it will bring a lot of copy-and-paste codes or we have to libraries common codes.

Instead, let me propose another approach.
Currently `tpm_state_path` in swtpm_setup is used in the following 2 ways.

- (1).  For `certsdir` which is used by swtpm_localca temporarily
- (2). For actual tpm state store path, ex:
  - check_state_overwride()
  - delete_swtpm_statefiles()
  - delete_state()

For 1. we can use temporal directory (using mkdtemp(3)) or a directory specified by user using new `--tempdir` (or `--certsdir`) option.
It will be prefereble for users who care sensitive data (EK) not written to disk. If swtpm_setup has such an option we can cover that requirement.

For 2. we can introduce new CMD to communicate with swtpm for example CMD_HAS_STATE for check_state_overwride(). Another idea is to introduce `--not-overwrite` in swtpm itself and pass the option from swtpm_setup then we can do equivalent tasks in swtpm like swtpm_setup does in check_state_overwride() and delete_swtpm_statefiles().
For delete_state(), I'm considering `--clear-state` for swtpm to remove states when init_tpm{,2} face some errors.

If we go this way, we can also reduce the swtpm_setup dependencies on swtpm implementation details (TPM state file names).

Any comments are appreciated.

stefanberger:
Now that you introduce a URI, I a wondering whether we should work with URIs like the iscsi:// you showed and now we would use a URI file://<path> or actually as mentioned in PR #513 maybe this should be the directory backend, rather than file, with a dir:// URI since this will want to own the whole directory due to the lock. Just to be consistent.

Etsukata:
> with a dir:// URI

Thanks, I like that idea. Will update this PR.

stefanberger:
> > with a dir:// URI
> 
> Thanks, I like that idea. Will update this PR.

That should then obsolete the `backend=` option parameter but this here should be enough: 

```
--tpmstate backend_uri=dir:/tmp/myvtpm
```

This assumes of course that every type of backend can be expressed via a URI, but as far as the current backend and the iscsi backends are concerned that should be sufficient.

stefanberger:
@Etsukata You will have to rebase this branch again. Also, if we don't end up merging it by Friday it will likely have to wait another 2 weeks or so since I will be out for a while.

Etsukata:
Hi @stefanberger, thanks for the notice.
I've rebased to master and removed `--tpmstate backend`.

man/man8/swtpm.pod
```diff
@@ -150,14 +150,18 @@ The following options are support by all interfaces:
 
 =over 4
 
-=item B<--tpmstate dir=E<lt>dirE<gt>[,mode=E<lt>0...E<gt>]>
+=item B<--tpmstate dir=E<lt>dirE<gt>[,mode=E<lt>0...E<gt>]|backend_uri=E<lt>uriE<gt>>
```
stefanberger, line 153: Hopefully last change. Please change this to `backend-uri` to be consistent with other option parameters.

stefanberger: The help screens in swtpm.c, swtpm_chardev.c, and cuse_tpm also need to be extended.

Etsukata: Sure, I'll update them.

man/man8/swtpm.pod
```diff
@@ -150,14 +150,18 @@ The following options are support by all interfaces:
...
 
-The TPM state files will be written with the given file mode bits.
-This value must be given as an octal number starting with a '0'.
+If I<dir> is specified, the TPM state files will be written to the I<dir> with
+the given file I<mode> bits. This value must be given as an octal number starting with a '0'.
 The default value is 0640.
 
+If I<backend_uri> is specified, the TPM state data will be stored to the URI.
+Currently I<backend_uri=dir://<path_to_dir>> is the only one available. In this case,
+URI should specify the path to the directory where files are stored.
```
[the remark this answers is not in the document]
Etsukata: Thanks, I will append the sentence to the man pod.

stefanberger:
@Etsukata I think it's ready to be merged...

pksnx (approved):
Thanks much for this, looks good to me.

Etsukata:
Hi @stefanberger, thanks for the comment.
I've added the following change to the first commit. I think it's good to merge.

```
diff --git a/src/swtpm/swtpm_nvstore_dir.h b/src/swtpm/swtpm_nvstore_dir.h
index 7cf467b..eec1dfd 100644
--- a/src/swtpm/swtpm_nvstore_dir.h
+++ b/src/swtpm/swtpm_nvstore_dir.h
@@ -36,8 +36,8 @@
 /* OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.                */
 /********************************************************************************/
 
-#ifndef _SWTPM_NVSTORE_FILE_H
-#define _SWTPM_NVSTORE_FILE_H
+#ifndef _SWTPM_NVSTORE_DIR_H
+#define _SWTPM_NVSTORE_DIR_H
 
 #include <libtpms/tpm_types.h>
 #include <libtpms/tpm_library.h>
@@ -75,4 +75,4 @@ SWTPM_NVRAM_DeleteName_Dir(uint32_t tpm_number,
 
 extern struct nvram_backend_ops nvram_dir_ops;
 
-#endif /* _SWTPM_NVSTORE_FILE_H */
+#endif /* _SWTPM_NVSTORE_DIR_H */
```

Etsukata:
For swtpm_setup, I'll create several PRs separately. I appreciate if you could take a look at them once you get back, thanks.

ifnkhan:
Looks good to me @Etsukata. 
Thanks!

PR #491, opened 2021-07-20
Improvements to .gitignore

nchevsky:
- Add auto-generated files `debian/autoreconf.*`
- Remove `*.patch` (Quilt requires these under `debian/patches/`)

Avoids the following untracked files after a build:
```
Untracked files:
  (use "git add <file>..." to include in what will be committed)
        debian/autoreconf.after
        debian/autoreconf.before
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,7 +8,6 @@
 *.la
 *.lo
 *.o
-*.patch
 *~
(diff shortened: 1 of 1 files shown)


stefanberger:
Thanks.

PR #496, opened 2021-07-21
Remove obsolete manpage for swtpm_setup.sh(8)

nchevsky:
swtpm_setup.sh was replaced by swtpm_setup in v0.5.0.

This change was green-lighted in [this discussion](https://github.com/stefanberger/swtpm/commit/3d75c96537384eafd78943844cbc2179ac70fe65#commitcomment-53761589).


stefanberger:
Can you run `./configure` so you also get the change in `swtpm.spec`?

stefanberger:
LGTM.

nchevsky:
> Can you run `./configure` so you also get the change in `swtpm.spec`?

Done, but the reason `swtpm.spec` didn't make it into my original commit is that it's listed in `.gitignore` (as IMO it should be, like any other auto-generated file). We should consider either:

1. not shipping `swtpm.spec` at all (my recommendation, though I see it was [recently readded](https://github.com/stefanberger/swtpm/commit/7f1f1fff467fb33a1f183f369842b654dec4d678) deliberately), or
2. removing `swtpm.spec` from `.gitignore` if you really want to continue to ship this auto-generated file.

Let me know if you'd like me to do either of these in another pull request. 🙂

stefanberger:
I have used swtpm.spec to install dependencies without having to go through the cumbersome `./configure` to install dependencies so often that I thought it was a convenient script to have in the repo to get started quickly, at least on Fedora.

I would  then opt for 2).

PR #509, opened 2021-07-31
swtpm: Preserve main loop flags when handling CMD_SET_DATAFD

nchevsky:
The code that handles control channel command `CMD_SET_DATAFD`, introduced in [this commit](https://github.com/stefanberger/swtpm/commit/1fb8bb790f5bd7d81a97be7507d1063aa3bb2c58), overwrites `mlp->flags` instead of `OR`ing new bits onto it. This causes any flags previously set by the command-line argument parsers (in `swtpm::swtpm_main()` and `swtpm_chardev::swtpm_chardev_main()`) to be lost.

### DIFF
--- a/src/swtpm/ctrlchannel.c
+++ b/src/swtpm/ctrlchannel.c
@@ -790,7 +790,7 @@ int ctrlchannel_process_fd(int fd,
             goto err_bad_input;
         }
 
-        mlp->flags = MAIN_LOOP_FLAG_USE_FD | MAIN_LOOP_FLAG_KEEP_CONNECTION |
+        mlp->flags |= MAIN_LOOP_FLAG_USE_FD | MAIN_LOOP_FLAG_KEEP_CONNECTION |
                        MAIN_LOOP_FLAG_END_ON_HUP;
         mlp->fd = *data_fd;
 


stefanberger:
Can you write an example (or two) into the commit message of something that didn't work anymore due to this?

src/swtpm/ctrlchannel.c
```diff
@@ -790,7 +790,7 @@ int ctrlchannel_process_fd(int fd,
             goto err_bad_input;
         }
 
-        mlp->flags = MAIN_LOOP_FLAG_USE_FD | MAIN_LOOP_FLAG_KEEP_CONNECTION |
```
stefanberger, line 793: Can you prefix the title with 'swtpm:'.

stefanberger:
> 
> 
> Can you write an example (or two) into the commit message of something that didn't work anymore due to this?

Actually the available flags are:
```
#define MAIN_LOOP_FLAG_TERMINATE  (1 << 0)
#define MAIN_LOOP_FLAG_USE_FD     (1 << 1)
#define MAIN_LOOP_FLAG_KEEP_CONNECTION (1 << 2)
#define MAIN_LOOP_FLAG_END_ON_HUP (1 << 3)
```

The TERMINATE flag is the only one that was cleared previously, if it was set at all.  When swtpm is started by libvirt and QEMU talks to it I don't think that this flag was set. So I think this won't have any negative side effects. The MAIN_LOOP_FLAG_END_ON_HUP gets us out of the main loop then (comment there should include UnixIO...).

nchevsky:
> The TERMINATE flag is the only one that was cleared previously, if it was set at all. When swtpm is started by libvirt and QEMU talks to it I don't think that this flag was set. So I think this won't have any negative side effects. The MAIN_LOOP_FLAG_END_ON_HUP gets us out of the main loop then (comment there should include UnixIO...).

Yeah, I actually spent the better part of a day studying the main loop in different use cases (socket + TCP, socket + UnixIO, `CMD_SET_DATAFD`, etc.) and came to the same conclusion. Now, even though nothing is effectively broken by this at the moment, that's really just luck—I still consider this to be a bug and don't see a reason _not_ to fix it; plus, you never know when down the road you may add new flags that _will_ break expected behavior if overwritten. 🙂

For context, the reason I even discovered this was precisely the `MAIN_LOOP_FLAG_END_ON_HUP` flag being set in that same statement. I was testing `swtpm` with QEMU (manually; no `libvirt` involved) and, in spite of my _not_ providing the `--terminate` argument, `swtpm` would terminate every time the data connection dropped. I didn't want this—I wanted the daemon to stick around so that I could reuse it and start/stop my VM without having to relaunch `swtpm` along with it every time. I then realized this was happening for two reasons:

1. When QEMU exits gracefully, it issues `CMD_SHUTDOWN` to `swtpm`. This can't be changed, so let's forget about it for now.
2. Even if QEMU doesn't send `CMD_SHUTDOWN`, `swtpm` terminates anyway due to `MAIN_LOOP_FLAG_END_ON_HUP` having been set by the `CMD_SET_DATAFD` handler, which causes the main loop to terminate unconditionally when the data connection drops, regardless of whether `--terminate` was given by the user. In fact, the code that handles `--terminate` isn't even evaluated because the loops exits earlier.

To me, this is conceptually wrong because:

1. The manual page states that `--terminate` is supported by the `socket` interface, with no strings attached. However, when `CMD_SET_DATAFD` is involved, `--terminate` is bypassed and no longer respected (and this isn't documented).
2. The documentation implies that if one _doesn't_ specify `--terminate`, the daemon won't terminate, period. However, `CMD_SET_DATAFD` causes the daemon to terminate anyway, contrary to the documented behavior.
3. Conceptually, I see no reason why `CMD_SET_DATAFD` should even affect the daemon's behavior one way or the other. Why not just stick to doing what it says it does—setting the file descriptor—and nothing else?
4. There's no technical impediment to the daemon staying up when `CMD_SET_DATAFD` is involved; I've tested this (by removing the `MAIN_LOOP_FLAG_END_ON_HUP` flag and `SIGKILL`ing QEMU to avoid `CMD_SHUTDOWN`) and it works great. Most importantly, it allows me to choose the behavior I want by either giving or omitting `--terminate`. If QEMU front-ends (e.g. libvirt, Proxmox) want this behavior, they can (and should) just use `--terminate`.
4. While the documentation does state that `--fd` implies `--terminate`, no such statement is made for the case where a file descriptor is passed in via the control channel with `CMD_SET_DATAFD`. If this is in fact the intended behavior, we should at least document it.

Anyway, back to the original point: I feel we should change the `=` assignment to `|=` regardless of the above, even if it doesn't _currently_ fix anything—it's free and can only help avoid problems in the future.

stefanberger:
So maybe one solution would be to remove the MAIN_LOOP_FLAG_END_ON_HUP at the location of your patch. The  terminating behavior has been around for a long time, though. But at least QEMU ends up sending the SHUTDOWN command and QEMU is the main consumer for now, especially when it comes to passing of file descriptors, at least from what I know. I will have a look at this next week.

nchevsky:
> So maybe one solution would be to remove the MAIN_LOOP_FLAG_END_ON_HUP at the location of your patch.

Right, that's what I was getting at. Before submitting this PR, I spent a long time testing main loop corner cases with `MAIN_LOOP_FLAG_END_ON_HUP` removed and the `=` assignment changed to `|=` (to preserve `MAIN_LOOP_FLAG_TERMINATE` when and if set by the user). I found no regressions and my results are documented in [this spreadsheet](https://docs.google.com/spreadsheets/d/1Bp8LWRN15P1Y_sDpK1qXg6ZXG9IwnDoZ7zcZHRYZ98s/edit?usp=sharing).

I chose to initially only submit a fix for the `=` assignment because I expected some resistance to the other change (since it alters existing behavior), so I thought I'd start with something harmless and go from there. 🙂 If you're open to also removing the flag, I'm all for making both changes at once—it would certainly help justify the `|=` fix, since removing the flag without also fixing the assignment still leaves `--terminate` broken.

BTW, I'm also a Proxmox contributor and we're in the early stages of integrating `swtpm`. If we proceed with removing the flag, I'll make sure Proxmox VE uses `--terminate` and gets this implemented correctly from the start.

src/swtpm/ctrlchannel.c
stefanberger: I meant primarily the title of the commit ...

nchevsky: > I meant primarily the title of the commit ...

Yes, I was waiting until you decided what to do regarding the two changes (i.e. just one or both) so I could then amend the commit as needed. Would you like to merge this PR as-is with just the assignment fix, and follow up on the flag change separately?

stefanberger:
At last for QEMU we don't need --terminate from what I can see. The intended behavior is that swtpm terminates if QEMU abruptly terminates (ctrl-c on it) and that seems to be the case due to the control channel recognizing that the client disappeared.

https://github.com/stefanberger/swtpm/blob/master/src/swtpm/mainloop.c#L222-L229
```
           if (pollfds[CTRL_CLIENT_FD].revents & POLLIN) {
                ctrlclntfd = ctrlchannel_process_fd(ctrlclntfd,
                                                    &mainloop_terminate,
                                                    &locality, &tpm_running,
                                                    mlp);

                if (mainloop_terminate)
                    break;
            }
```

src/swtpm/ctrlchannel.c
stefanberger: yes, let's make this two PRs.

nchevsky:
Ctrl+C (`SIGINT`) is a graceful shutdown as far as QEMU is concerned, so it does indeed send `CMD_SHUTDOWN`, and `swtpm` terminates as expected. Nothing technically wrong there. (I do find it wrong for QEMU to unconditionally shut down `swtpm` for me without giving me a choice, but that's between QEMU and me.)

The code you pasted handles commands and sets `mainloop_terminate` when it receives `CMD_SHUTDOWN`, but it isn't involved in detecting connections that dropped for any other reason. The [spreadsheet](https://docs.google.com/spreadsheets/d/1Bp8LWRN15P1Y_sDpK1qXg6ZXG9IwnDoZ7zcZHRYZ98s/edit?usp=sharing) I previously linked shows where the main loop terminates when the data connection drops _not_ as the result of a command.

If you want to see what happens when QEMU _doesn't_ send `CMD_SHUTDOWN`, you have to kill it with `SIGKILL`. What you'll see then is that `swtpm` _still_ terminates, _even though you didn't specify_ `--terminate` ⚠️. It's terminating because of `MAIN_LOOP_FLAG_END_ON_HUP`, which is undocumented (and in my opinion incorrect) behavior. According to the documentation, when a connection simply drops (no command involved), the daemon should terminate if and only if `--terminate` was given. The `--terminate` flag and its documented behavior aren't being respected here, and that's all I'm trying to solve. Does that make sense?

When `swtpm` was integrated into QEMU, the designer(s) assumed that users would only want `swtpm` to launch and die along with QEMU (not what I personally want). This resulted in the following implementation decisions:

1. QEMU unconditionally sends `CMD_SHUTDOWN` on exit, and
2. `swtpm` sets `MAIN_LOOP_FLAG_END_ON_HUP` so it always dies along with QEMU, even if QEMU didn't send `CMD_SHUTDOWN`, and even if the user didn't use `--terminate` ⚠️.

Problems with the latter:

1. It's undocumented behavior.
2. It doesn't respect `--terminate` being unset.
3. The same behavior could've been achieved by invoking `swtpm` with `--terminate`. There was never any need for the `CMD_SET_DATAFD` handler to use `MAIN_LOOP_FLAG_END_ON_HUP`.

In an ideal world:

1. QEMU would have a switch to optionally suppress `CMD_SHUTDOWN`, so users could choose to leave their `swtpm` daemon alive. 👉🏻 I plan to contribute such a switch to QEMU.
2. `swtpm`'s `CMD_SET_DATAFD` handler wouldn't set `MAIN_LOOP_FLAG_END_ON_HUP`. 👉🏻 We can fix this now.
3. QEMU front-ends which want `swtpm` to always start and stop along with QEMU would invoke `swtpm` with `--terminate`, guaranteeing that behavior in a clean, predictable, documented way. 👉🏻 I can make sure Proxmox does this.
4. I could choose to run `swtpm` as an always-on system service, and have QEMU connect to and disconnect from it as many times as I wanted without anything killing the daemon without my say-so. 👉🏻 I'm already doing this with a patched QEMU (to suppress `CMD_SHUTDOWN`) and the `MAIN_LOOP_FLAG_END_ON_HUP` removed from `swtpm`'s `CMD_SET_DATAFD` handler.

stefanberger:
The --terminate flag would also have to be part of libvirt. Also, if libvirt has the capability to kill() QEMU and it would leave a running swtpm behind, then I don't think we can make this change so fast. We'd first have to add the --terminate flag to libvirt and see that that alone would terminate swtpm if the HUP flag wasn't set anymore and then only in a while remove the HUP flag. I don't want to see unused swtpm instances left behind.

stefanberger:
In libvirt there should be a `qemuExtDevicesStop()` after a termination of QEMU using `qemuProcessKill()`. I  only see this in [one place](https://github.com/libvirt/libvirt/blob/v7.6.0/src/qemu/qemu_process.c#L7853-L7859), but there are several places from which qemuProcessKill() is being called in that file. hm... 

So basically what we are loosing by removing the HUP flag is that swtpm is left running after an abnormal termination of QEMU (e.g. segfault) or a forceful termination of it (kill(SIGKILL)). The former shouldn't be an issue but the latter could be an issue. I haven't been able to recreate such a situation yet, but it's not clear under what conditions those various kill calls in qemu_process.c are being called...

Adding the `--terminate` flag to the command line does replace the potentially removed `HUP` flag (from ctrlchannel.c) seamlessly, though.

nchevsky:
> The --terminate flag would also have to be part of libvirt.

Right, as well as Proxmox VE and any QEMU front-end that wants `swtpm` to behave that way. Then again, `--terminate` was always around and did exactly what they wanted, so they should've just used `--terminate` instead of adding code that creates overlapping and undocumented behavior, overrides command-line arguments, and isn't controllable by the user. 🙂

> Also, if libvirt has the capability to kill(SIGKILL) QEMU and it would leave a running swtpm behind, then I don't think we can make this change so fast. We'd first have to add the --terminate flag to libvirt (...) and then only in a while remove the HUP flag.

Yes, of course. This is another reason I didn't initially include the flag removal in this PR: It requires coordination with third parties.

> So basically what we are loosing by removing the HUP flag is that swtpm is left running after an abnormal termination of QEMU (e.g. segfault) or a forceful termination of it (kill(SIGKILL)). The former shouldn't be an issue but the latter could be an 
issue. I haven't been able to recreate such a situation yet, but it's not clear under what conditions those various kill calls in qemu_process.c are being called...

Either one could be an issue:
- `SIGSEGV`: While rare, QEMU does crash sometimes, especially when using hardware passthrough.
- `SIGKILL`: Proxmox VE forcefully kills QEMU when VMs fail to stop after a long timeout. libvirt may do something similar.

So libvirt, Proxmox VE, and friends should definitely start using `--terminate`—it's always been the documented right way of achieving the behavior they want.

> Adding the --terminate flag to the command line does replace the potentially removed HUP flag (from ctrlchannel.c) seamlessly, though.

Yes, `--terminate` catches every possible case of the data connection dropping for any reason. `MAIN_LOOP_FLAG_END_ON_HUP` only catches a subset of cases, since `POLLHUP` doesn't apply to TCP (I know TCP is irrelevant to this discussion; I'm just saying 🙂).

zippy2:
> In libvirt there should be a `qemuExtDevicesStop()` after a termination of QEMU using `qemuProcessKill()`. I only see this in [one place](https://github.com/libvirt/libvirt/blob/v7.6.0/src/qemu/qemu_process.c#L7853-L7859), but there are several places from which qemuProcessKill() is being called in that file. hm...

That's perfectly fine. `qemuProcessKill()` is intended to just send a signal to QEMU process. It is `qemuProcessStop()` that is called unconditionally after QEMU process died (regardless of the case, i.e. even without libvirt sending any signal to it, e.g. on monitor HUP). Therefore, it is okay to call `qemuExtDevicesStop()` just from that one place. And there is no hotplug implemented for TPM devices either (which would require starting/killing `swtmp` accordingly) so we are good on this front too.

PR #513, opened 2021-08-05
swtpm: add "linear" nvram store backend (for single-file storage and block devices)

PiMaker:
This RFC is based on the work of @Etsukata et. al. in #461 and #490. It uses the proposed abstraction layer to implement another nvram storage backend.

The backend is called "linear" (which, IMO, is a not a good name [for the user facing side anyway], "file" would have been much better... but is already used by, well, the "file" directory backend - maybe we should call that one "directory" on the commandline?). It provides a second, lighter abstraction layer weakly based on the design proposed in #461 - the "linear" backend itself maps files into a linear address space, a "linear backend" is then used to write those changes to wherever they need to go.

One such "linear backend" is the "linear file" one implemented together with the abstraction in the first patch. It simply mmaps a file or block device and writes data to it, using ftruncate to manage its size automatically (in case it's a regular file, a block device must be preallocated beforehand).

There are several TODOs still open within the code, but it does pass the provided test cases, and I was able to attach a block-device based backing store to a QEMU VM and boot into Windows Server correctly with it.

Incomplete list of TODOs:
* locking (is it necessary?)
* actual URI parsing (currently, always "linear file" backend is chosen, without prefix or with "file://")
* currently a maximum of 15 states are allowed, maybe make this user configurable? would make the header size dynamic though...

I hope this work doesn't conflict with something else already in the works? If so, see this as another design draft. The mentioned user-space iSCSI layer could be put on top of my "linear" abstraction fairly easily too I'd imagine.

I honestly didn't plan to make this a fully functional implementation at first but then got carried away a bit - looking forward to further discussion :)

Thanks!

### DIFF
--- a/man/man8/swtpm.pod
+++ b/man/man8/swtpm.pod
@@ -159,10 +159,13 @@ the given file I<mode> bits. This value must be given as an octal number startin
 The default value is 0640.
 
 If I<backend-uri> is specified, the TPM state data will be stored to the URI.
-Currently I<backend-uri=dir://<path_to_dir>> is the only one available. In this case,
-URI should specify the path to the directory where files are stored. If I<path_to_dir>
-starts with a '/' then the path is interpreted as an absolute path, otherwise it is
-a path relative to the current directory.
+Currently I<backend-uri=dir://<path_to_dir>> and I<backend-uri=file://<path_to_dir>>
+are available. For 'dir://', the URI should specify the path to the directory where
+files are stored. If I<path_to_dir> starts with a '/' then the path is interpreted
+as an absolute path, otherwise it is a path relative to the current directory.
+For 'file://', the URI should specify a single file or block device where TPM state
+will be stored. A blockdevice must exist already and be big enough to store all
+state.
 
 =item B<--tpm2>
 
@@ -345,7 +348,8 @@ The I<--key> option supports the I<pwdfd=> parameter.
 
 =item B<nvram-backend-dir>
 
-The I<--tpmstate> option supports the I<backend-uri=dir://...> parameter.
+The I<--tpmstate> option supports the I<backend-uri=dir://...> and
+I<backend-uri=file://...> parameters.
 
 =item B<tpm-send-command-header>
 
--- a/man/man8/swtpm.pod
+++ b/man/man8/swtpm.pod
@@ -305,6 +305,7 @@ may contain the following:
         "cmdarg-key-fd",
         "cmdarg-pwd-fd",
         "nvram-backend-dir",
+        "nvram-backend-file",
         "tpm-send-command-header",
         "flags-opt-startup",
         "rsa-keysize-1024",
--- a/src/swtpm/swtpm_nvstore.c
+++ b/src/swtpm/swtpm_nvstore.c
@@ -180,6 +180,8 @@ TPM_RESULT SWTPM_NVRAM_Init(void)
     backend_uri = tpmstate_get_backend_uri();
     if (strncmp(backend_uri, "dir://", 6) == 0) {
         g_nvram_backend_ops = &nvram_dir_ops;
+    } else if (strncmp(backend_uri, "file://", 7) == 0) {
+        g_nvram_backend_ops = &nvram_linear_ops;
     } else {
         logprintf(STDERR_FILENO,
                   "SWTPM_NVRAM_Init: Unsupported backend.\n");
--- a/src/swtpm/swtpm_nvstore_linear.c
+++ b/src/swtpm/swtpm_nvstore_linear.c
@@ -1,0 +1,479 @@
+#include "config.h"
+
+#define _GNU_SOURCE
+#include <stdlib.h>
+#include <string.h>
+#include <endian.h>
+
+#include <libtpms/tpm_error.h>
+#include <libtpms/tpm_nvfilename.h>
+
+#include "swtpm.h"
+#include "swtpm_debug.h"
+#include "swtpm_nvstore.h"
+#include "swtpm_nvstore_linear.h"
+#include "logging.h"
+#include "utils.h"
+
+static struct {
+    TPM_BOOL      initialized;
+    char          *loaded_uri;
+    struct nvram_linear_store_ops *ops;
+
+    unsigned char *data;
+    uint32_t      length;
+    struct nvram_linear_hdr *hdr; /* points into *data */
+} state;
+
+/*
+    Attempts to flush the header of the linear state, if required by the store.
+*/
+static TPM_RESULT
+SWTPM_NVRAM_Linear_FlushHeader(const char* uri)
+{
+    if (state.ops->flush) {
+        return state.ops->flush(uri, 0, le16toh(state.hdr->hdrsize));
+    }
+    return 0;
+}
+
+/*
+    Attempts a resize and ensures that state is updated correctly and the given
+    new_size could actually be reached.
+*/
+static TPM_RESULT
+SWTPM_NVRAM_Linear_SafeResize(const char* uri, uint32_t new_size)
+{
+    TPM_RESULT rc = 0;
+    uint32_t result;
+
+    if (!state.ops->resize) {
+        return new_size < state.length ? 0 : TPM_SIZE;
+    }
+
+    rc = state.ops->resize(uri, &state.data, &result, new_size);
+    if (rc) {
+        return rc;
+    }
+
+    /* base address might have changed, update pointers */
+    state.hdr = (struct nvram_linear_hdr*)state.data;
+    state.length = result;
+
+    if (result < new_size) {
+        return TPM_SIZE;
+    }
+
+    return rc;
+}
+
+#define FILE_NR_INVALID 0xffffffff
+
+/*
+    Returns the offset into the state.hdr.files array given a TPM state name and
+    number. Will be FILE_NR_INVALID if out of bounds or unknown name.
+*/
+static uint32_t
+SWTPM_NVRAM_Linear_GetFileNr(const char *name)
+{
+    uint32_t rc = 0;
+    if (strcmp(name, TPM_PERMANENT_ALL_NAME) == 0) {
+        rc += 0;
+    } else if (strcmp(name, TPM_VOLATILESTATE_NAME) == 0) {
+        rc += 1;
+    } else if (strcmp(name, TPM_SAVESTATE_NAME) == 0) {
+        rc += 2;
+    } else {
+        logprintf(STDERR_FILENO,
+                  "SWTPM_NVRAM_Linear_GetFileOffset: Unknown name '%s'\n",
+                  name);
+        return FILE_NR_INVALID;
+    }
+    if (rc >= SWTPM_NVSTORE_LINEAR_MAX_STATES) {
+        logprintf(STDERR_FILENO,
+                  "SWTPM_NVRAM_Linear_GetFileOffset: File limit exceeded: %d\n",
+                  rc);
+        return FILE_NR_INVALID;
+    }
+    return rc;
+}
+
+/*
+    Allocate a new file entry in the linear address space of state.data.
+    The new file will be placed at the end.
+
+    Importantly, this may perform a resize, so pointers into state.data or
+    state.hdr must not be kept over this function call.
+*/
+static TPM_RESULT
+SWTPM_NVRAM_Linear_AllocFile(const char *uri, uint32_t file_nr, uint32_t size)
+{
+    TPM_RESULT rc = 0;
+    struct nvram_linear_hdr_file *file;
+    uint32_t new_offset = le16toh(state.hdr->hdrsize);
+    uint32_t new_size;
+    uint32_t cur_end;
+    uint32_t i;
+    uint32_t section_size = size;
+    ROUND_TO_NEXT_POWER_OF_2_32(section_size);
+
+    /* find end of current last file */
+    for (i = 0; i < SWTPM_NVSTORE_LINEAR_MAX_STATES; i++) {
+        file = &state.hdr->files[i];
+        if (!file->offset) {
+            continue;
+        }
+
+        cur_end = le32toh(file->offset) + le32toh(file->section_length);
+        if (cur_end > new_offset) {
+            new_offset = cur_end;
+        }
+    }
+
+    new_size = new_offset + section_size;
+    rc = SWTPM_NVRAM_Linear_SafeResize(uri, new_size);
+    if (rc) {
+        return rc;
+    }
+
+    file = &state.hdr->files[file_nr];
+    file->section_length = htole32(section_size);
+    file->data_length = htole32(size);
+    file->offset = htole32(new_offset);
+
+    TPM_DEBUG("SWTPM_NVRAM_Linear_AllocFile: allocated file %d @ %d "
+              "(len=%d section=%d)\n",
+              file_nr, new_offset, size, section_size);
+
+    return rc;
+}
+
+/*
+    Deallocate a file from state.data. It's entry in state.hdr will be zeroed,
+    and the file length adjusted accordingly (if 'resize' is TRUE).
+    If the file was not at the end, any following files will be moved forward,
+    as to not leave any holes. This simplifies the allocator strategy, since it
+    allows new files to always be placed at the end.
+
+    If resize is true, this may perform a resize, so pointers into state.data or
+    state.hdr must not be kept over this function call.
+*/
+static TPM_RESULT
+SWTPM_NVRAM_Linear_RemoveFile(const char *uri,
+                              uint32_t file_nr,
+                              TPM_BOOL resize)
+{
+    TPM_RESULT rc = 0;
+    uint32_t next_offset = 0xffffffff;
+    uint32_t state_end = 0;
+    uint32_t new_len;
+    uint32_t i, cur_offset, cur_end;
+    struct nvram_linear_hdr_file *file;
+    struct nvram_linear_hdr_file old_file = state.hdr->files[file_nr];
+
+    if (old_file.offset == 0) {
+        return 0;
+    }
+
+    TPM_DEBUG("SWTPM_NVRAM_Linear_RemoveFile: removing filenr %d (resize=%d)\n",
+              file_nr, resize);
+
+    state.hdr->files[file_nr].offset = 0;
+    state.hdr->files[file_nr].data_length = 0;
+    state.hdr->files[file_nr].section_length = 0;
+
+    /* find offset of file right after the one we remove, and adjust offsets */
+    for (i = 0; i < SWTPM_NVSTORE_LINEAR_MAX_STATES; i++) {
+        file = &state.hdr->files[i];
+        if (!file->offset) {
+            continue;
+        }
+
+        cur_offset = le32toh(file->offset);
+        if (cur_offset > le32toh(old_file.offset)) {
+            if (cur_offset < next_offset) {
+                next_offset = cur_offset;
+            }
+            cur_end = cur_offset + le32toh(file->section_length);
+            if (cur_end > state_end) {
+                state_end = cur_end;
+            }
+            file->offset = htole32(cur_offset -
+                                   le32toh(old_file.section_length));
+        }
+    }
+
+    if (next_offset != 0xffffffff) {
+        TPM_DEBUG("SWTPM_NVRAM_Linear_RemoveFile: compacting\n");
+        /* if we weren't the end, compact by moving following files forward */
+        memmove(state.data + le32toh(old_file.offset),
+                state.data + next_offset,
+                state_end - next_offset);
+    }
+
+    if (resize) {
+        new_len = state.length - le32toh(old_file.section_length);
+        rc = SWTPM_NVRAM_Linear_SafeResize(uri, new_len);
+        if (rc == 0) {
+            state.length = new_len;
+        }
+    }
+
+    return rc;
+}
+
+static TPM_RESULT
+SWTPM_NVRAM_Prepare_Linear(const char *uri)
+{
+    TPM_RESULT rc = 0;
+
+    TPM_DEBUG("SWTPM_NVRAM_Prepare_Linear: uri='%s'\n", uri);
+
+    if (state.initialized) {
+        if (strcmp(state.loaded_uri, uri) == 0) {
+            /* same URI loaded, this is okay, nothing to be done */
+            return 0;
+        }
+
+        logprintf(STDERR_FILENO,
+                  "SWTPM_NVRAM_PrepareLinear: Cannot prepare twice\n");
+        return TPM_FAIL;
+    }
+
+    state.loaded_uri = malloc(strlen(uri) + 1);
+    strcpy(state.loaded_uri, uri);
+
+    /* TODO: Parse URI prefixes ("iscsi://", "rbd://", etc...) */
+    state.ops = &nvram_linear_file_ops;
+
+    if ((rc = state.ops->open(uri, &state.data, &state.length))) {
+        return rc;
+    }
+
+    state.hdr = (struct nvram_linear_hdr*)state.data;
+
+    if (le64toh(state.hdr->magic) != SWTPM_NVSTORE_LINEAR_MAGIC) {
+        logprintf(STDOUT_FILENO,
+                  "Formatting '%s' as new linear NVRAM store\n",
+                  uri);
+
+        state.hdr->magic = htole64(SWTPM_NVSTORE_LINEAR_MAGIC);
+        state.hdr->version = SWTPM_NVSTORE_LINEAR_VERSION;
+        state.hdr->hdrsize = htole16(sizeof(struct nvram_linear_hdr));
+        memset(&state.hdr->files, 0, sizeof(state.hdr->files));
+
+        SWTPM_NVRAM_Linear_FlushHeader(uri);
+
+    } else {
+        /* assume valid state found */
+        if (state.hdr->version > SWTPM_NVSTORE_LINEAR_VERSION) {
+            logprintf(STDERR_FILENO,
+                      "SWTPM_NVRAM_PrepareLinear: Unknown format version: %d\n",
+                      state.hdr->version);
+            return TPM_FAIL;
+        }
+    }
+
+    state.initialized = TRUE;
+    return rc;
+}
+
+static TPM_RESULT
+SWTPM_NVRAM_LoadData_Linear(unsigned char **data,
+                            uint32_t *length,
+                            uint32_t tpm_number,
+                            const char *name,
+                            const char *uri)
+{
+    uint32_t file_nr;
+    uint32_t file_offset;
+    uint32_t file_data_len;
+    struct nvram_linear_hdr_file *file;
+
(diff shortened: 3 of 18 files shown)


stefanberger:
So where should this particular backend be used? What are the use cases versus the current one?

Also, do you intend to write multiple swtpms' states into this linear backend 'somehow' or what is the advantage of this backend versus the one that we currently have? I understand it works with block devices but just for making sense of using a block device one would have to write multiple swtpm instances' state into that block device 'somehow' otherwise using a block device seems wasteful. You would have to have some sort of directory and a unique identifier for each swtpm to be able to differentiate state blobs. How much space would you pre-allocate for each blob and what if some day that blob isn't big enough anymore, how do  you re-organize this? You need locking to possibly avoid concurrency issues at least when mucking with the directory. At least locking which would be easier if you are just on one host versus some sort of shared block device accesses by different hosts. You could use some sort of file for locking against concurrent access. Further, somehow you need to manage deletion of swtpm instances along with their blobs. This is currently a directory removal done by libvirt for example... None of this seems so convincing to me and the questions I have for iscsi are applying to a block device as well.

stefanberger:
So where should this particular backend be used? What are the use cases versus the current one?

Also, do you intend to write multiple swtpms' states into this linear backend 'somehow' or what is the advantage of this backend versus the one that we currently have? I understand it works with block devices but just for making sense of using a block device one would have to write multiple swtpm instances' state into that block device 'somehow' otherwise using a block device seems wasteful. You would have to have some sort of directory and a unique identifier for each swtpm to be able to differentiate state blobs. How much space would you pre-allocate for each blob and what if some day that blob isn't big enough anymore, how do you re-organize this? You need locking to possibly avoid concurrency issues at least when modifying the directory. At least locking which would be easier if you are just on one host versus some sort of shared block device accesses by different hosts. You could use some sort of file for locking against concurrent access. Further, somehow you need to manage deletion of swtpm instances along with their blobs. This is currently a directory removal done by libvirt for example... None of this seems so convincing to me and the questions I have for iscsi are applying to a block device as well.

PiMaker:
> So where should this particular backend be used? What are the use cases versus the current one?

The use cases are as described in #461. For VM management software, it is very much preferable to store state in a single file, regardless if that is a block device or a regular file (fortunately linux abstracts that away for us anyway). Our use case includes a High Availability stack based on block devices, and (as also described in #461) using FUSE or some other way of putting a filesystem on it just isn't the way forward.

The proposed iSCSI backend could also make use of the infrastructure added with this patch, so implementing it should be simpler (as long as the interface I propose is deemed compatible enough, open for changes of course).

> Also, do you intend to write multiple swtpms' states into this linear backend 'somehow' or what is the advantage of this backend versus the one that we currently have?

"The one we currently have" is the one using a directory on a filesystem? Well, then the advantage is that this one doesn't require that and works on a block device, yes ;)

> I understand it works with block devices but just for making sense of using a block device one would have to write multiple swtpm instances' state into that block device 'somehow' otherwise using a block device seems wasteful. You would have to have some sort of directory and a unique identifier for each swtpm to be able to differentiate state blobs. How much space would you pre-allocate for each blob and what if some day that blob isn't big enough anymore, how do you re-organize this?

That is already implemented in this draft. See `SWTPM_NVRAM_Linear_AllocFile` and `SWTPM_NVRAM_Linear_RemoveFile` in `src/swtpm/swtpm_nvstore_linear.c` from my first commit.

The way I'm doing it now is by sizing the allocation dynamically. A static header structure (`nvram_linear_hdr`, currently with a limit of 15 files, so 5 TPMs with up to 3 states each) contains pointers to where the data is located within the address space of the file. This isn't the most efficient way, as it requires potentially expensive moves to clean up any "holes" that open up on removal or resizing (which happens by removal and re-insertion), but since data sizes are fairly small, this shouldn't be much of an issue IMHO. This means however, that there is no maximum size of blobs, as they can grow dynamically (regular files are `ftruncate`d if too small, block devices must simply be big enough to begin with, that is the responsibility of the user).

This is the main reason for the second abstraction layer, the allocation strategy is unified, and it ensures compatibility between different implementations. A user could, for example, simply copy the data byte-for-byte from a block dev or file to an iSCSI target, change the backend URI, and swtpm would happily load the state.

> You need locking to possibly avoid concurrency issues at least when modifying the directory. At least locking which would be easier if you are just on one host versus some sort of shared block device accesses by different hosts. You could use some sort of file for locking against concurrent access.

Of course locking across multiple machines or similar distributed systems is out of scope entirely here. Locking on a single machine can be done, for example by creating a simple lock file based on the full path or major/minor combo in case of a block device. Then just `flock` away.

Our particular use case does not require locking, as we do that in our HA stack already, but I can implement something sensible if deemed a requirement.

> None of this seems so convincing to me and the questions I have for iscsi are applying to a block device as well.

As already said, storing state in a single file as opposed to a directory is a lot easier to handle for VM management. Be that via iSCSI, RBD, a block device, or whatever else you can come up with. swtpm is a supplement to QEMU, and there is a reason QEMU supports so many storage backends.

In our use case this is particularly true, but I can imagine other systems will see it as a benefit too. I believe this request was bound to come up sooner or later, as more hypervisor-management platforms add full TPM support to their stack (now that Windows is so set on requiring it).

PiMaker:
rebased for your convenience :)

PiMaker:
Any updates on this? I've rebased again and tried to increase test coverage by also testing on a block device, although the CI seems to "SKIP" both new tests now - it ran them for the first version of this PR, at least according to coveralls...

stefanberger:
So this PR is now beyond draft level? I will play around  with it a bit...

stefanberger:
It would certainly be desirable to have better coverage ...

Does this backend support concurrent access by multiple swtpms to the same device?

PiMaker:
> So this PR is now beyond draft level? I will play around with it a bit...

I think I misinterpreted what "Draft" means here, so I removed the tag. The code is only updated to the rebase, not fundamentally changed from the first version, which already worked fine.

> It would certainly be desirable to have better coverage ...

As said, the tests are there (`test_tpm2_save_load_state_2_linear` and `test_tpm2_save_load_state_2_block`) and in the first version the former was run, but this time around CI appears to have skipped both?

> Does this backend support concurrent access by multiple swtpms to the same device?

If by "device" you mean the backing file, then no, and since there is no locking atm (see my other posts) using it concurrently will lead to data corruption.

src/swtpm/swtpm_nvstore.c
```diff
@@ -199,6 +203,7 @@ void SWTPM_NVRAM_Shutdown(void)
         g_nvram_backend_ops->cleanup();
     memset(&filekey, 0, sizeof(filekey));
     memset(&migrationkey, 0, sizeof(migrationkey));
+    SWTPM_NVRAM_LinearFile_Cleanup();
```
stefanberger, line 206: This should be called via the .cleanup interface function rather than a direct call.

src/swtpm/swtpm_nvstore_linear.c
```diff
@@ -0,0 +1,427 @@
...
+#include "swtpm_nvstore_linear.h"
+#include "swtpm_nvstore_linear_file.h"
+#include "logging.h"
+#include "utils.h"
+
+struct nvram_backend_ops nvram_linear_ops = {
+    .prepare = SWTPM_NVRAM_Prepare_Linear,
+    .load    = SWTPM_NVRAM_LoadData_Linear,
+    .store   = SWTPM_NVRAM_StoreData_Linear,
+    .delete  = SWTPM_NVRAM_DeleteName_Linear,
```
stefanberger, line 23: .cleanup goes here

stefanberger: One more general comment. The dir:// backend locks other swtpm instances out from concurrently accessing the directory. I think this one should do that also or does it maybe support multilple swtpm instances accessing a single block device/file concurrently ?

stefanberger:
What is the advantage of this backend over the existing one? Is it performance? Why do we need it and who should make use of it?

I also think we should wait until we can setup a TPM instance using swtpm_setup where the backend type parameter needs to be passed from the swtpm_setup command line to the swtpm command line.

src/swtpm/swtpm_nvstore.c
PiMaker: makes sense, missed that interface

PiMaker:
> What is the advantage of this backend over the existing one? Is it performance? Why do we need it and who should make use of it?

I think my post from Aug 9 answers that already, but in short, it's about manageability. Most VM HA and management stacks I know (that includes ours) work with block devices rather than filesystems.

> I also think we should wait until we can setup a TPM instance using swtpm_setup where the backend type parameter needs to be passed from the swtpm_setup command line to the swtpm command line.

Sounds reasonable, is there work being done on this already?

Locking is something I can implement if necessary, but wouldn't be required for our use-case. It's not too trivial, since we can't just place a lock file next to the file (we might not own the surrounding dir), but somewhere in 'temp' mapped to the canonical filename should be ok.

stefanberger:
> > I also think we should wait until we can setup a TPM instance using swtpm_setup where the backend type parameter needs to be passed from the swtpm_setup command line to the swtpm command line.
> 
> Sounds reasonable, is there work being done on this already?

I haven't looked at. Maybe @Etsukata has something already?

tests/test_tpm2_save_load_state_2_block
```diff
@@ -0,0 +1,30 @@
...
+	rm -rf $STATEDIR
+	if [ -n "$STATEFILE" ]; then
+		losetup -d $STATEFILE
+	fi
+}
+
+# allocate 4 MiB file
+fallocate -x -l $((4 * 1024 * 1024)) "$STATEIMG"
+# and loop mount it
+STATEFILE=$(losetup --show -f $STATEIMG)
```
stefanberger, line 24: Can this not work on plain files? Using files would not require root rights.

src/swtpm/swtpm_nvstore_linear.h
```diff
@@ -0,0 +1,107 @@
...
+                             const char *name,
+                             const char *uri);
+
+TPM_RESULT
+SWTPM_NVRAM_DeleteName_Linear(uint32_t tpm_number,
+                              const char *name,
+                              TPM_BOOL mustExist,
+                              const char *uri);
+
+extern struct nvram_backend_ops nvram_linear_ops;
```
stefanberger, line 105: I suppose most of the structures are private can be moved into the .c file except for this last line here that we may sooner or later have to consolidated into a common .h file. See my cleanup patch in PR #556.

src/swtpm/swtpm_nvstore_linear_file.c
```diff
@@ -0,0 +1,274 @@
...
+    Provides a linear backend based on memory-mapping a filesystem path.
+    Can be used to access regular files (with automatic resizing) or block
+    devices (with pre-allocated constant size, must be big enough for all state,
+    otherwise writing will fail).
+*/
+struct nvram_linear_store_ops nvram_linear_file_ops = {
+    .open = SWTPM_NVRAM_LinearFile_Open,
+    .flush = SWTPM_NVRAM_LinearFile_Flush,
+    .resize = SWTPM_NVRAM_LinearFile_Resize,
+};
```
stefanberger, line 32: this should go to the bottom of the file and all functions it uses should become static and the header file can then go.

tests/test_tpm2_save_load_state_2_block
stefanberger: .... otherwise it could work with loopback mounted files when root rights are available, with plain files otherwise, or both cases and skipping the loopback testing when no root rights are available ?

Etsukata:
> > > I also think we should wait until we can setup a TPM instance using swtpm_setup where the backend type parameter needs to be passed from the swtpm_setup command line to the swtpm command line.
> > 
> > 
> > Sounds reasonable, is there work being done on this already?
> 
> I haven't looked at. Maybe @Etsukata has something already?

I haven't coded that part yet.
I really appreciate it if you could work on it @PiMaker, thanks.

tests/test_tpm2_save_load_state_2_block
PiMaker: `test_tpm2_save_load_state_2_block` tests block devices via loopdev (or skips if root is unavailable) and `test_tpm2_save_load_state_2_file` tests on regular files, even without root. This way, both code paths are exercised.

PiMaker:
> > > > I also think we should wait until we can setup a TPM instance using swtpm_setup where the backend type parameter needs to be passed from the swtpm_setup command line to the swtpm command line.
> > > 
> > > 
> > > Sounds reasonable, is there work being done on this already?
> > 
> > 
> > I haven't looked at. Maybe @Etsukata has something already?
> 
> I haven't coded that part yet.
> I really appreciate it if you could work on it @PiMaker, thanks.

I'll take a look after cleaning up and rebasing this PR.

PiMaker:
Changes:
* rebase
* use `cleanup` method for linear backend, add `cleanup` API to the store backends too and use it
* remove `swtpm_nvstore_linear_file.h` in favor of static declarations and put "store ops" struct def in `swtpm_nvstore_linear.h` instead, while `nvram_linear_ops` is moved to `swtpm_nvstore.h` like in #556
* cleaned up `swtpm_nvstore_linear.h` a bit, left `nvram_linear_hdr` and consorts in though, since it is used by `swtpm_nvstore_linear_file.c`
* consolidate entire "linear backend" state into single anonymous "state" struct instead of partially in a struct and partially just statics

PiMaker:
rebased again and added support for the new `check_state` function

stefanberger:
> edit: is the slight decrease in coverage ok for merging still? most of it comes from not testing the specific syscall error paths AFAICT

That would be ok, but what would be more important is that swtpm_setup can actually use this backend and we don't have support for this yet.

stefanberger:
... and you will be maintaining this backend long-term?

PiMaker:
see also: #570

This isn't that much code, so I'm sure either me or someone else from our team can step up if issues arise in the future.

src/swtpm/swtpm_nvstore_linear.c
```diff
@@ -0,0 +1,470 @@
...
+static TPM_RESULT
+SWTPM_NVRAM_Linear_SafeResize(const char* uri, uint32_t new_size)
+{
+    TPM_RESULT rc = 0;
+
+    if (!state.ops->resize) {
+        return new_size < state.length ? 0 : TPM_SIZE;
+    }
+
+    uint32_t result;
```
stefanberger, line 53: Please move this to the top.

src/swtpm/swtpm_nvstore_linear.c
```diff
@@ -0,0 +1,470 @@
...
+        /* shouldn't happen, but just to be safe */
+        return TPM_FAIL;
+    }
+
+    /*
+        TODO: at the moment, callers require a pointer that can be 'free'd, but
+        for efficiency, it would be better to return the mapped area directly
+    */
+    *length = file_data_len;
+    *data = malloc(*length);
```
stefanberger, line 319: Check for NULL pointer.

src/swtpm/swtpm_nvstore_linear_file.c
```diff
@@ -0,0 +1,274 @@
...
+}
+
+/*
+    Strip leading "file://" from uri if given.
+*/
+static const char*
+SWTPM_NVRAM_LinearFile_UriToPath(const char* uri)
+{
+    const char *path = uri;
+    if (strncmp(uri, "file://", 7) == 0) {
```
stefanberger, line 133: empty line after declaration of vars.

src/swtpm/swtpm_nvstore_linear_file.c
```diff
@@ -0,0 +1,274 @@
...
+
+static TPM_RESULT
+SWTPM_NVRAM_LinearFile_Open(const char* uri,
+                            unsigned char **data,
+                            uint32_t *length)
+{
+    TPM_RESULT rc = 0;
+    const char *path = SWTPM_NVRAM_LinearFile_UriToPath(uri);
+
+    if (mmap_state.fd) {
```
stefanberger, line 147: fd = 0 being a valid fd: mmap_state.fd >= 0  ?

src/swtpm/swtpm_nvstore_linear_file.c
```diff
@@ -0,0 +1,274 @@
...
+                  strerror(errno));
+        rc = TPM_FAIL;
+    }
+
+    return rc;
+}
+
+static void SWTPM_NVRAM_LinearFile_Cleanup(void)
+{
+    if (mmap_state.fd) {
```
stefanberger, line 206: same comment: mmap_state.fd >= 0

I would just do 

```
if (mmap_state.fd < 0)
    return;
```

src/swtpm/swtpm_nvstore_linear_file.c
```diff
@@ -0,0 +1,274 @@
...
+
+static TPM_RESULT
+SWTPM_NVRAM_LinearFile_Resize(const char* uri,
+                              unsigned char **data,
+                              uint32_t *new_length,
+                              uint32_t requested_length)
+{
+    TPM_RESULT rc = 0;
+
+    if (!mmap_state.fd) {
```
stefanberger, line 222: mmap_state.fd < 0   ?

src/swtpm/swtpm_nvstore_linear_file.c
```diff
@@ -0,0 +1,274 @@
...
+            logprintf(STDERR_FILENO,
+                      "SWTPM_NVRAM_LinearFile_Resize: Error in munmap: %s\n",
+                      strerror(errno));
+            rc = TPM_FAIL;
+        }
+        /* only complain when ftruncate fails if growing was requested */
+        if (rc == 0 &&
+            ftruncate(mmap_state.fd, requested_length) &&
+            mmap_state.size < requested_length)
+        {
```
stefanberger, line 244: move this '{' to previous line.

stefanberger (changes requested):

src/swtpm/swtpm_nvstore_linear_file.c
PiMaker: This is correct, but the resolution would be wrong (since fd is initialized to 0, but nothing is mapped at that point). I added a simple boolean "mapped" field to the state to encode this instead of relying on some special value integer fd, I think that's nicer.

stefanberger (changes requested):

man/man8/swtpm.pod
```diff
@@ -159,10 +159,13 @@ the given file I<mode> bits. This value must be given as an octal number startin
 The default value is 0640.
 
 If I<backend-uri> is specified, the TPM state data will be stored to the URI.
-Currently I<backend-uri=dir://<path_to_dir>> is the only one available. In this case,
-URI should specify the path to the directory where files are stored. If I<path_to_dir>
-starts with a '/' then the path is interpreted as an absolute path, otherwise it is
-a path relative to the current directory.
+Currently I<backend-uri=dir://<path_to_dir>> and I<backend-uri=file://<path_to_dir>>
+are available. For 'dir://', URI should specify the path to the directory where
```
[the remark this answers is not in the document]
stefanberger: Rest looks good now.

src/swtpm/swtpm_nvstore_linear.c
```diff
@@ -0,0 +1,477 @@
...
+#define FILE_NR_INVALID 0xffffffff
+
+/*
+    Returns the offset into the state.hdr.files array given a TPM state name and
+    number. Will be FILE_NR_INVALID if out of bounds or unknown name.
+*/
+static uint32_t
+SWTPM_NVRAM_Linear_GetFileNr(const char *name, uint32_t tpm_number)
+{
+    uint32_t rc = tpm_number * 3;
```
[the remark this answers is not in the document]
PiMaker: What's that for then? Should I just remove the parameter?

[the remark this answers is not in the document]
stefanberger: It was a parameter that we used in a multi-instance vTPM implementation but this parameter is not used anymore. Instead we are starting a vTPM instance for every VM. You can replace it with a hardcoded '0'.

stefanberger:
> This isn't that much code, so I'm sure either me or someone else from our team can step up if issues arise in the future.

Then I'd merge it but please remove the [RFC] from the PR if you really want this.

src/swtpm/swtpm_nvstore_linear.c
```diff
@@ -0,0 +1,476 @@
...
+            /* same URI loaded, this is okay, nothing to be done */
+            return 0;
+        }
+
+        logprintf(STDERR_FILENO,
+                  "SWTPM_NVRAM_PrepareLinear: Cannot prepare twice\n");
+        return TPM_FAIL;
+    }
+
+    state.loaded_uri = malloc(strlen(uri) + 1);
```
stefanberger, line 243: Is there a corresponding free() to this malloc() somewhere? It should probably be in the cleanup() function.

PiMaker:
> > This isn't that much code, so I'm sure either me or someone else from our team can step up if issues arise in the future.
> 
> Then I'd merge it but please remove the [RFC] from the PR if you really want this.

Thanks a lot! I just fixed up the last `malloc/free` comment. I didn't know you could edit the title of PRs on GH, would have done so sooner...

stefanberger:
Another issue cropped up running the tests locally as root:

```
[...]
FAIL: test_tpm2_save_load_state_2_block
PASS: test_tpm2_setbuffersize
PASS: test_tpm2_volatilestate
PASS: test_tpm2_wrongorder
PASS: test_tpm2_save_load_state_2_linear
PASS: test_tpm2_probe
PASS: test_tpm2_save_load_state_2
PASS: test_tpm2_swtpm_setup_overwrite
PASS: test_tpm2_save_load_state_3
```

My guess is the tests  test_tpm2_save_load_state_2 and its 2 new siblings now need 3 ports for 3 concurrent runs.

stefanberger:
The ports range of 65460-65469 is completely unused. Maybe we move the test there.
Another choice would be to rename `test_tpm2_save_load_state_2` to `test_tpm2_save_load_state_2.test` and then introduce dependencies via log files to serialize the test runs like here. All 3 tests would then have to have the suffix `.test` in their name.

# Prevent concurrency between these two
test_tpm2_samples_create_tpmca.log: test_tpm2_swtpm_localca_pkcs11.log

PiMaker:
Never had that here, I've ran the tests quite rigurously. I'd be fine with either change (other port or serializing make) if that fixes the issue for you.

Unfortunately I have to run for today (timezones...), but I can take a look tomorrow if you want. Otherwise I'd say make that a followup since you seem to have a fix aleady? :)

stefanberger:
> Unfortunately I have to run for today (timezones...), but I can take a look tomorrow if you want. Otherwise I'd say make that a followup since you seem to have a fix already? :)

There's time tomorrow also. I can reliably reproduce this error with this command line:

```
sudo bash -c "SWTPM_TEST_IBMTSS2=1 LSAN_OPTIONS=fast_unwind_on_malloc=0 make -j12 check VERBOSE=1"
```

I don't understand why ASAN didn't show the memory leak...

PiMaker:
> The ports range of 65460-65469 is completely unused. Maybe we move the test there.

That's what I went with, I moved all three `test_tpm2_save_load_state_2*` tests to that range and made `_block` and `_linear` differ. This fixes the tests with the reproducing command line on my end.

> I don't understand why ASAN didn't show the memory leak...

I don't have experience with ASAN, but weirdly enough, valgrind doesn't show any leakage either (if I revert the 'free' change in cleanup)...

stefanberger:
You seem to have built valgrind into the patch now. I get an error now in all 3 tests due to open file descriptor:

```
FAIL: test_tpm2_save_load_state_2
=================================

/usr/bin/tssnvdefinespace
/usr/bin/tssstartup
==2971800== Memcheck, a memory error detector
==2971800== Copyright (C) 2002-2017, and GNU GPL'd, by Julian Seward et al.
==2971800== Using Valgrind-3.17.0 and LibVEX; rerun with -h for copyright info
==2971800== Command: swtpm socket --print-capabilities
==2971800==
==2971800==
==2971800== FILE DESCRIPTORS: 3 open (3 std) at exit.
==2971800==
==2971800== HEAP SUMMARY:
==2971800==     in use at exit: 18,812 bytes in 10 blocks
==2971800==   total heap usage: 17 allocs, 7 frees, 23,601 bytes allocated
==2971800==
==2971800== LEAK SUMMARY:
==2971800==    definitely lost: 0 bytes in 0 blocks
==2971800==    indirectly lost: 0 bytes in 0 blocks
==2971800==      possibly lost: 0 bytes in 0 blocks
==2971800==    still reachable: 18,812 bytes in 10 blocks
==2971800==         suppressed: 0 bytes in 0 blocks
==2971800== Reachable blocks (those to which a pointer was found) are not shown.
==2971800== To see them, rerun with: --leak-check=full --show-leak-kinds=all
==2971800==
==2971800== For lists of detected and suppressed errors, rerun with: -s
==2971800== ERROR SUMMARY: 0 errors from 0 contexts (suppressed: 0 from 0)
==2975669== Memcheck, a memory error detector
==2975669== Copyright (C) 2002-2017, and GNU GPL'd, by Julian Seward et al.
==2975669== Using Valgrind-3.17.0 and LibVEX; rerun with -h for copyright info
==2975669== Command: swtpm socket --server port=65460 --tpmstate dir=/tmp/tmp.eokZYmcnwU --pid file=/tmp/tmp.eokZYmcnwU/swtpm.pid --ctrl type=unixio,path=/tmp/tmp.eokZYmcnwU/sock --log file=/tmp/tmp.eokZYmcnwU/logfile,level=20 --tpm2
==2975669==
Error: (1) Socket TPM did not write pidfile.
swtpm: Could not open UnixIO socket: No such file or directory
==2975669==
==2975669== FILE DESCRIPTORS: 3 open (3 std) at exit.
==2975669==
==2975669== HEAP SUMMARY:
==2975669==     in use at exit: 18,804 bytes in 9 blocks
==2975669==   total heap usage: 18 allocs, 9 frees, 19,131 bytes allocated
==2975669==
==2975669== LEAK SUMMARY:
==2975669==    definitely lost: 0 bytes in 0 blocks
==2975669==    indirectly lost: 0 bytes in 0 blocks
==2975669==      possibly lost: 0 bytes in 0 blocks
==2975669==    still reachable: 18,804 bytes in 9 blocks
==2975669==         suppressed: 0 bytes in 0 blocks
==2975669== Reachable blocks (those to which a pointer was found) are not shown.
==2975669== To see them, rerun with: --leak-check=full --show-leak-kinds=all
==2975669==
==2975669== For lists of detected and suppressed errors, rerun with: -s
==2975669== ERROR SUMMARY: 0 errors from 0 contexts (suppressed: 0 from 0)
```

stefanberger:
When I remove the added line with valgrind it runs for me.

PiMaker:
... sorry, clearly my mistake. That wasn't meant to go to git. The 3 fd "leaks" are fine though, they correspond to stdin/stdout/stderr. *Hopefully* last fixup incoming.

stefanberger:
> _Hopefully_ last fixup incoming.

It looks good now. I will merge it once Travis is done.

stefanberger (approved):

PR #514, opened 2021-08-06
Check libtpms version

### LINKED ISSUE #512: [RFE] Offer configure-time option to disable TPM 1.2 support
Red Hat is interested in removing TPM 1.2 support from its build of swtpm.

To make that possible without nasty downstream-only hacks, ideally swtpm would gain a `--without-tpm12` or similar configure switch, which results in all code related to TPM 1.2 being compiled out; it would also start reporting whether TPM 1.2 support is present in the binary through its existing introspection facilities, so that libvirt can detect when it's missing and warn users who attempt to use it.

Does this sound like a reasonable feature to add? Do you foresee any big obstacle to implementing it? Assuming the answers are yes and no respectively, @elmarco and I can take a stab at it :)

elmarco:
After libtpms gains support for disabling TPM 1.2 (https://github.com/stefanberger/libtpms/pull/244), adapt swtpm to cope with and check the libtpms supported versions.

I left a couple of things for later (eh, someone else? :)):
- Some tests could be improved to work regardless of the supported versions.
- Also, it may be worth to make the test check the version at runtime, to avoid having to run configure again when libtpms is modified.

Resolves #512

### DIFF
--- a/man/man8/swtpm.pod
+++ b/man/man8/swtpm.pod
@@ -290,6 +290,8 @@ may contain the following:
     {
       "type": "swtpm",
       "features": [
+        "tpm-1.2",
+        "tpm-2.0",
         "cmdarg-seccomp",
         "cmdarg-key-fd",
         "cmdarg-pwd-fd",
@@ -298,12 +300,29 @@ may contain the following:
         "rsa-keysize-1024",
         "rsa-keysize-2048",
         "rsa-keysize-3072"
-      ]
+      ],
+      "version": "0.7.0"
     }
+
+The version field is available since 0.7.
 
 The meaning of the feature verbs is as follows:
 
 =over 4
+
+=item B<tpm-1.2>
+
+TPM 1.2 emulation is supported (libtpms is compiled with 1.2 support).
+
+Since 0.7
+
+==item B<tpm-2.0>
+
+TPM 2 emulation is supported (libtpms is compiled with 2.0 support).
+
+(the I<--tpm2> option is supported)
+
+Since 0.7
 
 =item B<cmdarg-seccomp>
 
--- a/src/swtpm/capabilities.c
+++ b/src/swtpm/capabilities.c
@@ -43,6 +43,7 @@
 #include <string.h>
 
 #include <libtpms/tpm_library.h>
+#include <libtpms/tpm_error.h>
 
 #include "capabilities.h"
 #include "logging.h"
@@ -122,19 +123,28 @@ int capabilities_print_json(bool cusetpm)
 #else
     const char *cmdarg_seccomp = "";
 #endif
+    const char *with_tpm1 = "";
+    const char *with_tpm2 = "";
     char *keysizecaps = NULL;
 
     ret = get_rsa_keysize_caps(&keysizecaps);
     if (ret < 0)
         goto cleanup;
 
+    if (TPMLIB_ChooseTPMVersion(TPMLIB_TPM_VERSION_1_2) == TPM_SUCCESS)
+        with_tpm1 = "\"tpm-1.2\", ";
+    if (TPMLIB_ChooseTPMVersion(TPMLIB_TPM_VERSION_2) == TPM_SUCCESS)
+        with_tpm2 = "\"tpm-2.0\", ";
+
     n =  asprintf(&string,
          "{ "
          "\"type\": \"swtpm\", "
          "\"features\": [ "
-             "%s%s%s%s%s%s"
+             "%s%s%s%s%s%s%s%s"
           " ] "
          "}",
+         with_tpm1,
+         with_tpm2,
          !cusetpm     ? "\"tpm-send-command-header\", ": "",
          !cusetpm     ? "\"flags-opt-startup\", "      : "",
          cmdarg_seccomp,
--- a/src/swtpm/capabilities.c
+++ b/src/swtpm/capabilities.c
@@ -68,7 +68,7 @@ static int get_rsa_keysize_caps(char **keysizecaps)
     int n;
 
     if (!info_data)
-        goto oom;
+        goto cleanup;
 
     start = strstr(info_data, needle);
     if (start) {
--- a/src/swtpm/capabilities.c
+++ b/src/swtpm/capabilities.c
@@ -141,7 +141,8 @@ int capabilities_print_json(bool cusetpm)
          "\"type\": \"swtpm\", "
          "\"features\": [ "
              "%s%s%s%s%s%s%s%s"
-          " ] "
+          " ], "
+         "\"version\": \"" VERSION "\" "
          "}",
          with_tpm1,
          with_tpm2,
--- a/tests/common
+++ b/tests/common
@@ -831,3 +831,15 @@ function validate_pidfile()
 		exit 1
 	fi
 }
+
+# Check whether swtpm can use a TPM 1.2
+function skip_test_no_tpm12()
+{
+	local swtpm_exe="$1"
+
+	local res=$(${swtpm_exe} socket --print-capabilities | grep '"tpm-1.2"')
+	if [ -z "${res}" ]; then
+		echo "${swtpm_exe} does not provide a TPM 1.2"
+		exit 77
+	fi
+}
--- a/tests/common
+++ b/tests/common
@@ -843,3 +843,15 @@ function skip_test_no_tpm12()
 		exit 77
 	fi
 }
+
+# Check whether swtpm can use a TPM 2.0
+function skip_test_no_tpm20()
+{
+	local swtpm_exe="$1"
+
+	local res=$(${swtpm_exe} socket --print-capabilities | grep '"tpm-2.0"')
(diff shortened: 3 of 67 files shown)


src/swtpm/capabilities.c
```diff
@@ -122,19 +123,28 @@ int capabilities_print_json(bool cusetpm)
...
+    const char *with_tpm1 = "";
+    const char *with_tpm2 = "";
     char *keysizecaps = NULL;
 
     ret = get_rsa_keysize_caps(&keysizecaps);
     if (ret < 0)
         goto cleanup;
 
+    if (TPMLIB_ChooseTPMVersion(TPMLIB_TPM_VERSION_1_2) == TPM_SUCCESS)
+        with_tpm1 = "\"tpm-1.2\", ";
```
stefanberger, line 135: If this is now not showing tpm-1.2 then this would mean that no tpm-1.2 is available. Older versions of swtpm would not show that either, giving the impression that TPM 1.2 was not available. I think we need to reverse the logic here and show 'no-tpm-1.2' also because this is the new feature you added to libtpms.

src/swtpm/capabilities.c
```diff
@@ -122,19 +123,28 @@ int capabilities_print_json(bool cusetpm)
...
+    if (TPMLIB_ChooseTPMVersion(TPMLIB_TPM_VERSION_2) == TPM_SUCCESS)
+        with_tpm2 = "\"tpm-2.0\", ";
```
stefanberger, line 137: Same here, or we show both, no-tpm-2.0 and tpm-2.0, which would make it clear whether it is available or not.

configure.ac
```diff
@@ -190,6 +190,9 @@ PKG_CHECK_EXISTS([libtpms >= 0.9], [
     )
 ])
 
+AM_CONDITIONAL([WITH_TPM1], test "x$WITH_TPM1" != "x")
```
stefanberger, line 193: I saw your message in the PR. I think you should drop this patch. The following should resolve the issue but we'll need to add it to all test cases:

```iff --git a/tests/common b/tests/common
index 5abe951..ee675ef 100644
--- a/tests/common
+++ b/tests/common
@@ -831,3 +831,15 @@ function validate_pidfile()
                exit 1
        fi
 }
+
+# Check whether swtpm can use a TPM 1.2
+function skip_test_no_tpm12()
+{
+       local swtpm_exe="$1"
+
+       local res=$(${swtpm_exe} socket --print-capabilities | grep '"tpm-1.2"')
+       if [ -z "${res}" ]; then
+               echo "${swtpm_exe} cannot use a TPM 1.2"
+               exit 77
+       fi
+}
diff --git a/tests/test_commandline b/tests/test_commandline
index e311bf8..d1acc7a 100755
--- a/tests/test_commandline
+++ b/tests/test_commandline
@@ -14,6 +14,8 @@ TESTDIR=${abs_top_testdir:=$(dirname "$0")}
 # need SWTPM to be set
 source ${TESTDIR}/common
 
+skip_test_no_tpm12 "${SWTPM_EXE}"
+
 TPMDIR=`mktemp -d`
 PID_FILE=$TPMDIR/${SWTPM}.pid
 LOG_FILE=$TPMDIR/${SWTPM}.log
```

src/swtpm/capabilities.c
```diff
@@ -43,6 +43,7 @@
 #include <string.h>
 
 #include <libtpms/tpm_library.h>
+#include <libtpms/tpm_error.h>
```
stefanberger, line 46: Can you prefix the title of this commit with `swtpm:`

configure.ac
```diff
@@ -179,6 +179,17 @@ AC_CHECK_LIB(tpms,
 )
 AC_SUBST([LIBTPMS_LIBS])
 
+PKG_CHECK_EXISTS([libtpms >= 0.9], [
```
stefanberger, line 182: If you drop patch 4 this one is probably not necessary, either.

stefanberger (changes requested):

src/swtpm/capabilities.c
stefanberger: test_tpm2_print_capabilities and test_print_capabilities will need to be adjusted as well..

configure.ac
stefanberger: I have that patch now ...

src/swtpm/capabilities.c
elmarco: Why not version the capabilities? I think it it would be simpler. Introducing a negative capabilty is awkward and still something we cannot retrofit, so users will need to adjust depending on the version.

src/swtpm/capabilities.c
elmarco: I don't think it's necessary. (see my other comment)

configure.ac
elmarco: great, so I will pick your branch!

src/swtpm/capabilities.c
elmarco: sure

configure.ac
elmarco: right

tests/_test_encrypted_state
```diff
@@ -36,6 +36,7 @@ trap "cleanup" EXIT
 
 [ "${SWTPM_INTERFACE}" == cuse ] && source ${TESTDIR}/test_cuse
 source ${TESTDIR}/common
+skip_test_no_tpm12 "${SWTPM_EXE}"
```
stefanberger, line 39: I think what you did is better. Let me adjust.

stefanberger: I updated my branch now with a replacement patch.

src/swtpm/capabilities.c
stefanberger: Then let's try it with a version.

tests/common
```diff
@@ -843,3 +843,15 @@ function skip_test_no_tpm12()
 		exit 77
 	fi
 }
+
```
stefanberger, line 846: Please also prefix this patch with `tests:`.

tests/_test_encrypted_state
elmarco: But you dropped the 2.0 skips? Any of the two versions can be disabled, both are needed.

stefanberger: > 
> 
> But you dropped the 2.0 skips? Any of the two versions can be disabled, both are needed.

I only updated my patch ...

tests/common
elmarco: ok

tests/_test_encrypted_state
elmarco: Hmm, my patch for 2.0 doesn't work: 
common: line 852: /src/swtpm/swtpm: No such file or directory

elmarco: ah I see, I should update based on your changes. ok

src/swtpm/capabilities.c
elmarco: You are right though, if we want to backport this, we should have some "no-tpm". Tbh, I think we shouldn't backport such major change.

src/swtpm/capabilities.c
stefanberger: Can you please prefix the title of this patch with `swtpm:`.

stefanberger:
We also need an update to man/man8/swtpm.pod.

stefanberger (approved):

stefanberger:
Unfortunately we need yet more test cases patched. With this libtpms configure line:
```
./configure --prefix=/usr --without-tpm1 --without-tpm2  --with-openssl
```

And this command line to run the test cases:
```
sudo bash -c "SWTPM_TEST_EXPENSIVE=1 SWTPM_TEST_IBMTSS2=1 make -j32 check"
```

I get these failure cases:
```
FAIL: test_tpm2_vtpm_proxy
FAIL: test_tpm12
FAIL: test_tpm2_ibmtss2
FAIL: test_vtpm_proxy
```

This one also needs patching:
```
SKIP: test_tpm2_partial_reads
```

man/man8/swtpm.pod
```diff
@@ -298,13 +300,30 @@ may contain the following:
...
 
+=item B<tpm-1.2>
+
+TPM 1.2 emulation is supported (libtpms is compiled with 1.2 support).
+
+Since 0.7
+
+==item B<tpm-2.0>
+
+TPM 2.0 emulation is supported (libtpms is compiled with 2.0 support).
```
stefanberger, line 321: All other documentation says TPM 2. Please change this.

elmarco: done

src/swtpm/capabilities.c
stefanberger: I didn't mean backporting but interpretation of this new verb now depends on the version identifier. If tpm-1.2 is missing prior to version 0.7 then that doesn't mean that TPM 1.2 isn't there.

PR #515, opened 2021-08-09
swtpm_setup: Add support for checking for TPM 1.2 and TPM 2 support     

stefanberger:
Implement get_supported_tpm_versions to get swtpm's support for TPM 1.2
and TPM 2 and use it error out in case user choose a TPM version that
is not supported. Also display the supported TPM versions in the
capabilites JSON.

### DIFF
--- a/src/swtpm_setup/swtpm_setup.c
+++ b/src/swtpm_setup/swtpm_setup.c
@@ -1328,7 +1328,7 @@ int main(int argc, char *argv[])
     swtpm_prg_l = split_cmdline(swtpm_prg);
     tmp = g_find_program_in_path(swtpm_prg_l[0]);
     if (!tmp) {
-        logerr(gl_LOGFILE, "TPM at %s is not an executable.\n", swtpm_prg);
+        logerr(gl_LOGFILE, "swtpm at %s is not an executable.\n", swtpm_prg_l[0]);
         goto error;
     }
     g_free(tmp);
--- a/src/swtpm_setup/swtpm_setup.c
+++ b/src/swtpm_setup/swtpm_setup.c
@@ -879,6 +879,30 @@ static void usage(const char *prgname, const char *default_config_file)
         );
 }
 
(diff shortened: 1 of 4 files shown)


stefanberger:
This will have to wait until PR #514 has been applied.

src/swtpm_setup/swtpm_setup.c
```diff
@@ -879,6 +879,37 @@ static void usage(const char *prgname, const char *default_config_file)
...
+    int ret = 1;
+    g_autofree gchar **argv = NULL;
+    gchar *my_argv[] = { "--print-capabilities", NULL };
+
+    argv = concat_arrays(swtpm_prg_l, my_argv, FALSE);
+    success = g_spawn_sync(NULL, argv, NULL, G_SPAWN_STDERR_TO_DEV_NULL, NULL, NULL,
+                           &standard_output, NULL, &exit_status, &error);
+    if (!success) {
+        logerr(gl_LOGFILE, "Could not start swtpm '%s': %s\n", swtpm_prg_l[0], error->message);
+        g_error_free(error);
```
elmarco, line 898: Why not also use g_autoptr(GError) ?

elmarco: I would also drop the error label, and return directly.

stefanberger: Done.

src/swtpm_setup/swtpm_setup.c
```diff
@@ -879,6 +879,30 @@ static void usage(const char *prgname, const char *default_config_file)
         );
 }
 
+static int get_supported_tpm_versions(gchar **swtpm_prg_l, gboolean *swtpm_has_tpm12,
+                                      gboolean *swtpm_has_tpm2)
+{
+    gboolean success;
+    g_autofree gchar *standard_output = NULL;
+    int exit_status = 0;
```
elmarco, line 887: if you don't care about exist_status, you may just pass NULL

elmarco:
lgtm

PR #516, opened 2021-08-13
swtpm_setup: Use temp dir for certs instead of TPM state dir

Etsukata:
Currently swtpm_setup saves ek.cert and platform.cert to TPM state
directory and removes them later. As the usage of these files are
temporal, temp directory can be used for this purpose. This reduces
dependencies on TPM state dir, which is convenient for pluggable state
store (https://github.com/stefanberger/swtpm/issues/461).

### DIFF
--- a/src/swtpm_setup/swtpm_setup.c
+++ b/src/swtpm_setup/swtpm_setup.c
@@ -442,9 +442,9 @@ static int tpm2_create_eks_and_certs(unsigned long flags, const gchar *config_fi
 static int init_tpm2(unsigned long flags, gchar **swtpm_prg_l, const gchar *config_file,
                      const gchar *tpm2_state_path, const gchar *vmid, const gchar *pcr_banks,
                      const gchar *swtpm_keyopt, int *fds_to_pass, size_t n_fds_to_pass,
-                     unsigned int rsa_keysize, const gchar *user_certsdir)
+                     unsigned int rsa_keysize, const gchar *certsdir,
+                     const gchar *user_certsdir)
 {
-    g_autofree gchar *certsdir = g_strdup(tpm2_state_path);
     struct swtpm2 *swtpm2;
     struct swtpm *swtpm;
     int ret;
@@ -622,9 +622,9 @@ static int tpm12_create_certs(unsigned long flags, const gchar *config_file,
 static int init_tpm(unsigned long flags, gchar **swtpm_prg_l, const gchar *config_file,
                     const gchar *tpm_state_path, const gchar *ownerpass, const gchar *srkpass,
                     const gchar *vmid, const gchar *swtpm_keyopt,
-                    int *fds_to_pass, size_t n_fds_to_pass, const gchar *user_certsdir)
+                    int *fds_to_pass, size_t n_fds_to_pass, const gchar *certsdir,
+                    const gchar *user_certsdir)
 {
-    g_autofree gchar *certsdir = g_strdup(tpm_state_path);
     struct swtpm12 *swtpm12;
     struct swtpm *swtpm;
     g_autofree gchar *pubek = NULL;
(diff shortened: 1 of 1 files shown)


src/swtpm_setup/swtpm_setup.c
```diff
@@ -1187,6 +1188,7 @@ int main(int argc, char *argv[])
     time_t now;
     struct tm *tm;
     int ret = 1;
+    GError *err = NULL;
```
stefanberger, line 1191: ` g_autoptr(GError) error = NULL;`

Otherwise looks good.

src/swtpm_setup/swtpm_setup.c
```diff
@@ -1561,12 +1563,20 @@ int main(int argc, char *argv[])
           curr_grp ? curr_grp->gr_name : "<unknown>",
           tmpbuffer);
 
+    certsdir = g_dir_make_tmp("swtpm_setup.certs.XXXXXX", &err);
```
stefanberger, line 1566: I think we should be able to restrict the creation of the temp dir with
`if (flags & (SETUP_EK_CERT_F | SETUP_PLATFORM_CERT_F) != 0) {...`

src/swtpm_setup/swtpm_setup.c
Etsukata: Oops, thanks.

src/swtpm_setup/swtpm_setup.c
Etsukata: Thanks. Let me use `if (flags & (SETUP_EK_CERT_F | SETUP_PLATFORM_CERT_F)) {` because `tpm2_create_ek_and_cert()` is using it.

src/swtpm_setup/swtpm_setup.c
```diff
@@ -1561,12 +1563,22 @@ int main(int argc, char *argv[])
           curr_grp ? curr_grp->gr_name : "<unknown>",
           tmpbuffer);
 
+    if (flags & (SETUP_EK_CERT_F | SETUP_PLATFORM_CERT_F)) {
+        certsdir = g_dir_make_tmp("swtpm_setup.certs.XXXXXX", &error);
+        if (certsdir == NULL) {
+            logerr(gl_LOGFILE, "Could not create temporal directory for certs: %s\n",
```
stefanberger, line 1569: temporal -> temporary

src/swtpm_setup/swtpm_setup.c
```diff
@@ -1590,5 +1602,9 @@ int main(int argc, char *argv[])
     g_strfreev(swtpm_prg_l);
     g_free(gl_LOGFILE);
 
+    if (certsdir && g_rmdir(certsdir) != 0)
+        logerr(gl_LOGFILE, "Could not remove temporal directory for certs: %s\n",
```
stefanberger, line 1606: same here. will merge then.

Etsukata: Thanks will fix them and commit message as well.

src/swtpm_setup/swtpm_setup.c
```diff
@@ -1590,5 +1602,9 @@ int main(int argc, char *argv[])
     g_strfreev(swtpm_prg_l);
     g_free(gl_LOGFILE);
 
+    if (certsdir && g_rmdir(certsdir) != 0)
+        logerr(gl_LOGFILE, "Could not remove temporary directory for certs: %s\n",
```
pksnx, line 1606: `gl_LOGFILE` could be potentially freed (via line 1603), maybe move that free before the exit ?

Etsukata: Good catch. I'll move that line before frees.

pksnx: Thanks, looks good to me.

pksnx (approved):

PR #524, opened 2021-09-06
Introduce "--print-states" to check TPM state existence

Etsukata:
Currently, swtpm_setup checks TPM state existence by directly accessing TPM state dir. To support pluggable storage backend https://github.com/stefanberger/swtpm/issues/461 , it's better for swtpm_setup not to have dependencies on implementation details of TPM state backend.

This PR introduces a new option "--print-states" to swtpm. With this option, swtpm shows TPM state status as follows:

```
      $ swtpm socket --print-states --tpmstate dir=/tmp --tpm2 | jq .
      {
        "type": "swtpm",
        "states": [
          {
            "name": "tpm2-00.permall"
          }
        ]
      }
```

Through that output, swtpm_setup can check existence of TPM state.

Current "--print-states" only shows "permall" state as it's the only one currently required for swtpm_setup. Of course it's open to discuss whether we should have other names ("volatile", "savestate").

### Commits Summary

* 2386357 swtpm_setup: Add tests for --overwrite options
  * Tests for swtpm_setup to ensure we can keep current behavior.
* d79a52c swtpm: Add GetFilepathForName()
  * Move GetFilenameForName() to nvstore.c so that we can use the func from "--print-states"
* bd95b57 swtpm: Add --print-states for reporting TPM states status
  * Implementation of "--print-states".
* 268d4dd swtpm_setup: Use --print-states in check_state_overwrite()
  * Replace check_state_overwrite() implementation.

PR #526, opened 2021-09-10
Minor fixes for --print-states option

Etsukata:
Some minor fixes for `--print-states` which is recently introduced in https://github.com/stefanberger/swtpm/pull/524. This PR includes function name change (Print_Json => PrintJson), a test fix and error handling fix.


stefanberger:
Thanks. Looks good to me.

PR #527, opened 2021-09-14
configure: drop python dependency

baloo:
as far as I can tell, 0.6.0 dropped the python dependency, yet the
configure still looks out for a python to be available. I don't believe
it makes any use of it.

This reduces the dependency graph on build.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,8 +17,6 @@ addons:
         - sed
         - bash
         - dh-exec
-        - python3-pip
-        - python3-twisted
         - libfuse-dev
         - libglib2.0-dev
         - libjson-glib-dev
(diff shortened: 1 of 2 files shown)


stefanberger:
We have these testcases here that require python:

```
# ls tests/*.py
tests/test_clientfds.py  tests/test_setdatafd.py
```

We could get rid of the python-twisted dependency if we also remove this part here in `test_setdatafd.py`:

```
if sys.version_info[0] < 3:
    import twisted.python.sendmsg as sendmsg
```

This should be possible since we don't support python2 anymore.

baloo:
They require python, but, unless I miss something, they are not executed during the build at any point. They're only distributed as extra_dist as far as I can tell.

stefanberger:
These are test cases that run on `make check` from these tests:
```
tests/test_commandline:167:LOG=$(PID_FILE=$TPMDIR/swtpm.pid SWTPM_EXE=$SWTPM_EXE TPMDIR=$TPMDIR exec $TESTDIR/test_clientfds.py)
tests/test_ctrlchannel3:50:LOG=$(SOCK_PATH=$SWTPM_CTRL_UNIX_PATH exec $TESTDIR/test_setdatafd.py)
```

baloo:
Correct, I had an issue in my packaging which made the tests not to run. Sorry for the noise.

PR #531, opened 2021-09-16
tests: Fix test_tpm2_save_load_state_da_timeout rare timer race

Etsukata:
When there is a delay between "Sending TPM2_NV_Write with wrong
password" and "timenow=$(date +%s)", the test can fail with "Error: Did
not get expected failure from TPM2_NV_Write() with good password.
Lockout should be enabled.". This can sometimes happen on slow or busy
systems.

To stabilize the test, this patch introduces precise recovery time
measurement ($timerecovery) and check good password works after that
time. As for lockout timeout, moves starting timer just before
TPM2_NV_Write and check good password fails before that time.

### DIFF
--- a/tests/_test_tpm2_save_load_state_da_timeout
+++ b/tests/_test_tpm2_save_load_state_da_timeout
@@ -121,6 +121,8 @@ if [ "$RES" != "$exp" ]; then
 	exit 1
 fi
 
+timenow=$(date +%s)
+timeout=$((timenow + 6))
 # Send TPM2_NV_Write with wrong password
 swtpm_open_cmddev ${SWTPM_INTERFACE} 100
 RES=$(swtpm_cmd_tx ${SWTPM_INTERFACE} ${NVWRITE_BAD})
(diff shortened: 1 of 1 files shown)


Etsukata:
> When there is a delay between ...

We can easily reproduce this by inserting `sleep 1` as follows:
```
$ git diff
diff --git a/tests/_test_tpm2_save_load_state_da_timeout b/tests/_test_tpm2_save_load_state_da_timeout
index cd2fe47..9435a9a 100755
--- a/tests/_test_tpm2_save_load_state_da_timeout
+++ b/tests/_test_tpm2_save_load_state_da_timeout
@@ -131,6 +131,7 @@ if [ "$RES" != "$exp" ]; then
        echo "received: $RES"
        exit 1
 fi
+sleep 1
 timenow=$(date +%s)
 timeout=$((timenow + 6))
```

stefanberger:
I haven't seen issues with this test in ages, but thanks for looking into this.

Issue #538, opened 2021-09-18
Use not-OpenSSL-3.0.0-deprecated APIs in /src/swtpm/


stefanberger:
swpm_aes.c is using the following OpenSSL-3.0.0-deprecated  APIs:

```
swtpm_aes.c:128:13: error: 'AES_set_encrypt_key' is deprecated [-Werror,-Wdeprecated-declarations]
        if (AES_set_encrypt_key(tpm_symmetric_key_data->userKey,
[...]
swtpm_aes.c:143:9: error: 'AES_cbc_encrypt' is deprecated [-Werror,-Wdeprecated-declarations]
        AES_cbc_encrypt(decrypt_data_pad,
[...]
swtpm_aes.c:216:13: error: 'AES_set_decrypt_key' is deprecated [-Werror,-Wdeprecated-declarations]
        if (AES_set_decrypt_key(tpm_symmetric_key_data->userKey,
[...]
swtpm_aes.c:227:9: error: 'AES_cbc_encrypt' is deprecated [-Werror,-Wdeprecated-declarations]
        AES_cbc_encrypt(encrypt_data,

```

swtpm_nvstore.c is using the following OpenSSL-3.0.0-deprecated APIs:
```
swtpm_nvstore.c:514:21: error: 'HMAC_CTX_new' is deprecated [-Werror,-Wdeprecated-declarations]
    HMAC_CTX *ctx = HMAC_CTX_new();
[...]
swtpm_nvstore.c:521:10: error: 'HMAC_Init_ex' is deprecated [-Werror,-Wdeprecated-declarations]
    if (!HMAC_Init_ex(ctx, key, key_len, EVP_sha256(), NULL) ||
[...]
swtpm_nvstore.c:522:10: error: 'HMAC_Update' is deprecated [-Werror,-Wdeprecated-declarations]
        !HMAC_Update(ctx, in, in_length))
[...]
swtpm_nvstore.c:526:10: error: 'HMAC_Update' is deprecated [-Werror,-Wdeprecated-declarations]
        !HMAC_Update(ctx, ivec, ivec_length))
[...]
swtpm_nvstore.c:529:10: error: 'HMAC_Final' is deprecated [-Werror,-Wdeprecated-declarations]
    if (!HMAC_Final(ctx, md, md_len))
[...]
swtpm_nvstore.c:538:5: error: 'HMAC_CTX_free' is deprecated [-Werror,-Wdeprecated-declarations]
    HMAC_CTX_free(ctx);
```

PR #545, opened 2021-09-20
swtpm: Add implementation of SWTPM_HMAC using OpenSSL 3.0 APIs

stefanberger:
Add an implementation of SWTPM_HMAC using non-deprecated APIs when
compiling with OpenSSL 3.0.

This partially addresses issue #538.


stefanberger:
I tested this with ASAN and comparsion-checked results of new SWTPM_HMAC implementation versus old one by using them both and comparing results. All tests passed.

Issue #537, opened 2021-09-18
Use not-deprecated OpenSSL API in swtpm_setup


stefanberger:
swtpm_setup is using APIS deprected in OpenSSL 3.0.0:

```
swtpm.c:1548:19: error: 'RSA_new' is deprecated [-Werror,-Wdeprecated-declarations]
    RSA *rsakey = RSA_new();
[...]
swtpm.c:1589:9: error: 'RSA_set0_key' is deprecated [-Werror,-Wdeprecated-declarations]
    if (RSA_set0_key(rsakey, mod, exp, NULL) != 1) {
[...]
swtpm.c:1595:9: error: 'EVP_PKEY_assign' is deprecated [-Werror,-Wdeprecated-declarations]
        EVP_PKEY_assign_RSA(pkey, rsakey) != 1) {
[...]
swtpm.c:1719:5: error: 'RSA_free' is deprecated [-Werror,-Wdeprecated-declarations]
    RSA_free(rsakey);
[...]
```

stefanberger:
The branch https://github.com/stefanberger/swtpm/tree/master.ossl3.rsa has a prototype for resolving this issue but it currently doesn't work due to `EVP_PKEY_encrypt_init(ctx)` failing since `ctx->pkey = NULL`.

Trying to get some help from the OpenSSL community: https://github.com/openssl/openssl/issues/16635#issuecomment-923066393

stefanberger:
Also here: https://github.com/openssl/openssl/issues/16639

PR #550, opened 2021-09-21
Use OpenSSL 3.0 APIs for RSA key creation

stefanberger:
Use OpenSSL 3.0 APIs for RSA key creation. This fixes issue #537.


stefanberger:
I tested this with ASAN and `tests/test_parameters` is also using it quite extensively, so it should be correct.

PR #553, opened 2021-09-22
aes: Avoid defining an already defined macro

tbaederr:
printf() is already defined as a macro when the compiler in use does not
support __va_arg_pack, which is the case for e.g. clang:
```
../../../src/swtpm/swtpm_aes.c:55:9: error: 'printf' macro redefined
[-Werror,-Wmacro-redefined]
        ^
/usr/include/bits/stdio2.h:115:11: note: previous definition is here
          ^
1 error generated.
```

### DIFF
--- a/src/swtpm/swtpm_aes.c
+++ b/src/swtpm/swtpm_aes.c
@@ -52,7 +52,9 @@
 #include "swtpm_aes.h"
 #include "logging.h"
 
+#ifndef printf
 #define printf(X ...)
+#endif
 
 typedef const EVP_CIPHER *(*evpfunc)(void);
 


stefanberger:
Is it defined to something that will generate output because that's what we are trying to avoid? Should it maybe rather be the following?

```
#ifdef printf
#undef printf
#endif
```

stefanberger:
The best is probably to convert some of the printf's for error messages to logprintf's and get rid of the rest.

stefanberger:
I now opened PR #555 to get rid of the printfs. Thanks for reporting.

tbaederr:
I only thought about `#undef`'ing after I'd already opened the PR. Thanks for fixing it.

PR #554, opened 2021-09-22
swtpm_setup: Forward --logfile option to swtpm

Etsukata:
Whenever swtpm_setup is executed with --logfile option, forward the option to swtpm (--log file=...). This helps debugging swtpm initialization issues.

### DIFF
--- a/src/swtpm/swtpm.c
+++ b/src/swtpm/swtpm.c
@@ -406,6 +406,15 @@ int swtpm_main(int argc, char **argv, const char *prgname, const char *iface)
         exit(EXIT_FAILURE);
     }
 
+    /* change process ownership before accessing files */
+    if (runas) {
+        if (change_process_owner(runas) < 0)
+            exit(EXIT_FAILURE);
+    }
+
+    if (handle_log_options(logdata) < 0)
+        exit(EXIT_FAILURE);
+
     if (mlp.fd >= 0 && mlp.fd < 3) {
         /* no std{in,out,err} */
         logprintf(STDERR_FILENO,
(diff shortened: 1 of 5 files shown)


src/swtpm/swtpm.c
```diff
@@ -406,6 +406,9 @@ int swtpm_main(int argc, char **argv, const char *prgname, const char *iface)
         exit(EXIT_FAILURE);
     }
 
+    if (handle_log_options(logdata) < 0)
+        exit(EXIT_FAILURE);
+
```
stefanberger, line 411: This has now been moved before this part here, which could now lead to permission issues when accessing existing log files.
```
    /* change process ownership before accessing files */
    if (runas) {
        if (change_process_owner(runas) < 0)
            goto exit_failure;
    }
```

Etsukata: Thanks, I missed that. I'll move `change_process_owner()` before `handle_log_options()`.

stefanberger: Tested this and it looks good.

stefanberger:
There's now unfortunately another issue... If the state of the vTPM is encrypted and one wants to run swtpm_setup again on it using `--overwrite` then this command line spawned by swtpm_setup fails:
```
/usr/bin/swtpm socket --print-states --tpmstate dir=/tmp/myvtpm --tpm2
```

It needs to `--key` option to be passed like in `src/swtpm/swtpm_setup` line 90:

```
   if (self->keyopts != NULL) {
        keyopts = g_strdup(self->keyopts);
        argv = concat_arrays(argv, (gchar*[]){"--key", keyopts, NULL}, TRUE);
    }
```

stefanberger:
Maybe the backend interface needs to have a function to detect availability of states so that `int SWTPM_NVRAM_PrintJson(void)` can then ask the backend whether the type of state is available and in the case of the dir backend it would become a simple `stat()` on the path and file?

Etsukata:
> it would become a simple stat() on the path and file?

I agree with your idea and checked https://github.com/stefanberger/swtpm/pull/556. Having explicit CheckState is easier to understand.
Thanks for the fix.

stefanberger:
Please rebase it on master.

stefanberger:
Can you please rebase on master when you get a chance?

Etsukata:
> Can you please rebase on master when you get a chance?

Sure, rebased.

PR #556, opened 2021-09-24
swtpm: Implement CheckState interface function

stefanberger:
This PR does some cleanup on the directory backend and implements a CheckState interface function for checking for availability of the permanent state blob on the backend.

@Etsukata  Please have a look.


src/swtpm/swtpm_nvstore.c
```diff
@@ -1324,10 +1324,10 @@ int SWTPM_NVRAM_PrintJson(void)
     TPM_RESULT rc = 0;
     int ret = 0, n;
     unsigned char *nvdata = NULL;
```
Etsukata, line 1326: `nvdata` is no longer used. We can remove this.

Etsukata (approved):
Thanks @stefanberger, I put a cleanup comment (`nvdata`). Other than that, it looks good to me.

src/swtpm/swtpm_nvstore.c
stefanberger: Thanks. Hadn't looked. Updated.

stefanberger:
> Thanks @stefanberger, I put a cleanup comment (`nvdata`). Other than that, it looks good to me.

Before it was also only looking for availability of the file, so now we have swtpm pretty much do the same and not try out the file.

Etsukata (approved):
Thanks, looks good to me.

PR #557, opened 2021-09-25
swtpm: Search for all state files and use abstracted names in JSON

stefanberger:
Search for all the state files not just the permanent state and
when printing the JSON use the abstracted names rather than concrete
filenames that are only valid for the dir backend but will likely
not exist in other backends.

Adjust swtpm_setup to search for the abstracted name and also
adjust the error message to print out the abstracted name.

Adjust the test cases.


stefanberger:
This is a follow-up to PR #556 .

@Etsukata

Etsukata (approved):

PR #558, opened 2021-09-26
Tests: Add a test case for testing --runas and ability to access files

stefanberger:
Add a test case for testing the --runas parameter and the ability to access
existing files that have restrictive access permission so that only allow
the --runas user is able to access these files. This is to ensure that
swtpm can access these files properly when the handling of options and the
changing to the --runas user is moved around within swtpm.

### DIFF
--- a/tests/_test_tpm2_file_permissions
+++ b/tests/_test_tpm2_file_permissions
@@ -1,0 +1,255 @@
+#!/bin/bash
+
+# For the license, see the LICENSE file in the root directory.
+#set -x
+
+ROOT=${abs_top_builddir:-$(pwd)/..}
+TESTDIR=${abs_top_testdir:-$(dirname "$0")}
+
+VTPM_NAME="vtpm-test-file-permissions"
+SWTPM_DEV_NAME="/dev/${VTPM_NAME}"
+export TPM_PATH=$(mktemp -d)
+STATE_FILE=${TPM_PATH}/tpm2-00.permall
+VOLATILE_STATE_FILE=${TPM_PATH}/tpm2-00.volatilestate
+PIDFILE=${TPM_PATH}/swtpm.pid
+LOGFILE=${TPM_PATH}/swtpm.log
+PWDFILE=${TPM_PATH}/pwdfile.txt
+SWTPM_INTERFACE=${SWTPM_INTERFACE:-cuse}
+SWTPM_SETUP_CONF=${TPM_PATH}/swtpm_setup.conf
+
+function cleanup()
+{
+	if [ -n "${SWTPM_PID}" ]; then
+		kill_quiet -9 "${SWTPM_PID}"
+	fi
+	rm -rf "${TPM_PATH}"
+}
+
+trap "cleanup" EXIT
+
+[ "${SWTPM_INTERFACE}" == cuse ] && source ${TESTDIR}/test_cuse
+source ${TESTDIR}/common
+
+cat <<_EOF_ > "${SWTPM_SETUP_CONF}"
+create_certs_tool=unused
+create_certs_tool_config=/dev/null
+create_certs_tool_options=/dev/null
+_EOF_
+
+# We need to copy swtpm and libswtpm_libtpms.so to the workdir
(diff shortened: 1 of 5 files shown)


stefanberger:
@Etsukata I would like to slide this test in before your changes so that we can test the before and after.

tests/_test_tpm2_file_permissions
```diff
@@ -0,0 +1,174 @@
...
+	exit 1
+fi
+
+if wait_process_gone ${SWTPM_PID} 4; then
+	echo "Error: ${SWTPM_INTERFACE} TPM should not be running anymore."
+	exit 1
+fi
+
+if [ -f "${PIDFILE}" ]; then
+	echo "Error: PID file shouldld have been removed."
```
Etsukata, line 132: should

tests/_test_tpm2_file_permissions
```diff
@@ -0,0 +1,174 @@
...
+
+logsize=$(get_filesize "${LOGFILE}")
+
+${SWTPM_SETUP} \
+	--tpmstate "${TPM_PATH}" \
+	--tpm2 \
+	--logfile "${LOGFILE}" \
+	--runas ${TESTUSER}
+if [ $? -eq 0 ]; then
+	echo "Error: ${SWTPM_SETUP} shoudl have refused to overwrite existing state."
```
Etsukata, line 147: should

tests/_test_tpm2_file_permissions
```diff
@@ -0,0 +1,174 @@
...
+# Init the TPM
+run_swtpm_ioctl ${SWTPM_INTERFACE} -i
+if [ $? -ne 0 ]; then
+	echo "Error: Could not initialize the ${SWTPM_INTERFACE} TPM."
+	exit 1
+fi
+
+# Volatile state file must be gone now
+if [ -f "${VOLATILE_STATE_FILE}" ]; then
+	echo "Error: Volatile state file has not be removed."
```
Etsukata, line 67: be -> been

Etsukata (commented):
> @Etsukata I would like to slide this test in before your changes so that we can test the before and after.

Sounds good! Thanks for adding this test.

> This is to ensure that
swtpm can access these files properly when the handling of options and the
changing to the --runas user is moved around within swtpm.

In this test, `run_swtpm` is executed as `root` user so even if we move `handle_log_option()` as follows (before `change_process_ownership` ), this test won't fail.

```
diff --git a/src/swtpm/swtpm.c b/src/swtpm/swtpm.c
index e1161b6..964ca8d 100644
--- a/src/swtpm/swtpm.c
+++ b/src/swtpm/swtpm.c
@@ -406,6 +406,10 @@ int swtpm_main(int argc, char **argv, const char *prgname, const char *iface)
         exit(EXIT_FAILURE);
     }
 
+    if (handle_log_options(logdata) < 0) {
+        exit(EXIT_FAILURE);
+    }
+
     if (mlp.fd >= 0 && mlp.fd < 3) {
         /* no std{in,out,err} */
         logprintf(STDERR_FILENO,
@@ -457,8 +461,7 @@ int swtpm_main(int argc, char **argv, const char *prgname, const char *iface)
             goto exit_failure;
     }
 
-    if (handle_log_options(logdata) < 0 ||
-        handle_key_options(keydata) < 0 ||
+    if (handle_key_options(keydata) < 0 ||
         handle_migration_key_options(migkeydata) < 0 ||
         handle_pid_options(piddata) < 0 ||
         handle_locality_options(localitydata, &mlp.locality_flags) < 0 ||
```

I'm now considering how to catch this.

stefanberger:
In this test case I am reading the state files directly. Maybe we should use the --printstates options for that an reintroduce the {'name': 'permall', 'size': 1072} type of enumeration with sizes added to be able to abstract this test case bit. I am not sure whether it will be useful for someone to know not just whether a type of state exists but also what the size is of the blob. Hm. @Etsukata ?

Etsukata:
> type of enumeration with sizes added to be able to abstract this test case bit. 

Sounds good to me. I've checked https://github.com/stefanberger/swtpm/pull/563.

PR #562, opened 2021-09-27
debian: Remove trousers from list of dependencies

stefanberger:
Trouser is needed by samples/swtpm-create-tpmca which in turn needs
GnuTLS tpmtool but doesn't install on Ubuntu machines when there's no
TPM 1.2 available or /dev/tpm0 is a TPM 2, so remove it from the list
of dependencies.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -57,13 +57,12 @@ Architecture: any
 Description: Tools for the TPM emulator
  The swtpm-tools package contains the following types of tools:
   - swtpm_bios: Tool for initializing the TPM
-  - swtpm_ioctl: Tool for controlling the CUSE TPM
+  - swtpm_ioctl: Tool for controlling the CUSE/socket/chardev TPM
   - swtpm_setup: Tool for creating the initial state of the TPM; this
       tool basically simulates TPM manufacturing where certificates are
       written into the NVRAM of the TPM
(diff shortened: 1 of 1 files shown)


nchevsky:
We may have some code here and there that, while not actually using TrouSerS, does have an unconditional dependency on its `tss` user. For example:

```
debian/swtpm-tools.postinst:  chown tss:root $SWTPM_LOCALCA_DIR
                                    ^^^
```

There may be more; this was just off the top of my head.

Speaking of `tpmtool(1)`, Debian stubbornly builds their `gnutls-bin` package with TPM support unconditionally turned off ([Debian bug](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=790500), [Ubuntu bug](https://bugs.launchpad.net/ubuntu/+source/gnutls28/+bug/1702755)) even though upstream (GnuTLS) has had TPM support enabled by default for _seven years_.

I submitted [this merge request](https://salsa.debian.org/gnutls-team/gnutls/-/merge_requests/2) to the Debian maintainer to try to get the `--without-tpm` dropped from the build (lots of background information in the merge request description and comments), but he doesn't seem interested.

stefanberger:
Another choice for use would be to create the tss user and group in a swtpm post install script file if that user/group don't exist.

PR #563, opened 2021-09-27
swtpm: Add size to each type of state and use JSON object

stefanberger:
Add the size of the type state to the --print-states output and switch
back to a JSON object when enumerating the blobs.


stefanberger:
@Etsukata I am switching back to the JSON object in the state type enumeration. Since the lock is locked I cannot actually use the --print-states to display the sizes of the state files while the swtpm is running, but I think it's better to have the JSON object and also the size in it.

PR #570, opened 2021-09-30
swtpm_setup: improve --tpmstate handling to allow support for file:// backends

PiMaker:
Most of the work is done abstracting away the specific implementation for working with directories in `swtpm_setup.c`. Passing the parameter itself forward is simple enough.

This PR first implements the abstraction layer for the `dir://` backend, with full backwards compatibility (no functional change intended). Afterwards, a new implementation for `file://` is added.

Includes a test case. Also tested in our environment directly.

This is of course based on my other PR #513 , so the added test case might fail initially. Any other test cases should not fail, as the old interface is fully supported still.


src/swtpm_setup/swtpm.c
```diff
@@ -67,16 +67,19 @@ static int swtpm_start(struct swtpm *self)
...
     unsigned ctr;
+    int pidfile_fd;
+    int ret = 1;
 
-    self->pidfile = g_strjoin(G_DIR_SEPARATOR_S,
-                              self->state_path, ".swtpm_setup.pidfile", NULL);
-    pidfile_file = g_strdup_printf("file=%s", self->pidfile);
+    g_free(self->pidfile);
+    self->pidfile = g_strdup(".swtpm_setup.pidfile.XXXXXX");
+    pidfile_fd = mkstemp(self->pidfile);
```
stefanberger, line 75: Do we still need `self->pidfile` at all then since this is now just a template.

stefanberger: Na, you are right. It will get the actual filename then.

src/swtpm_setup/swtpm.c
```diff
@@ -67,16 +67,19 @@ static int swtpm_start(struct swtpm *self)
     gboolean success;
     GError *error = NULL;
     unsigned ctr;
+    int pidfile_fd;
+    int ret = 1;
 
-    self->pidfile = g_strjoin(G_DIR_SEPARATOR_S,
-                              self->state_path, ".swtpm_setup.pidfile", NULL);
-    pidfile_file = g_strdup_printf("file=%s", self->pidfile);
+    g_free(self->pidfile);
```
stefanberger, line 73: I don't think this should be necessary since we aren't trying to re-start the program in case it fails, or did we have a memory leak here before?

src/swtpm_setup/swtpm.c
PiMaker: Hm, I kept it so it can be unlinked in 'cleanup', but that could be done locally in the function as well (since the fd is already open)... I can change it if you want.

src/swtpm_setup/swtpm.c
```diff
@@ -132,26 +130,29 @@ static int swtpm_start(struct swtpm *self)
...
-            return 0;
+            ret = 0;
+            break;
         }
         usleep(5000);
     }
 
-    return 1;
+error:
+    close(pidfile_fd);
```
stefanberger, line 154: This function should remove the pidfile if swtpm isn't running once it returns.

src/swtpm_setup/swtpm_backend_dir.c
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    g_dir_close(dir);
+
+    return ret;
+}
+
+/* Free an instance of dir_state. */
+static void free_dir_state(void *state) {
+    g_free(state);
```
stefanberger, line 100: No `g_free(state->dir)`?

src/swtpm_setup/swtpm_backend_dir.c
```diff
@@ -0,0 +1,108 @@
...
+                break;
+            }
+        }
+        if (g_str_has_suffix(fn, "permall") ||
+            g_str_has_suffix(fn, "volatilestate") ||
+            g_str_has_suffix(fn, "savestate")) {
+            g_autofree gchar *fullname = g_strjoin(G_DIR_SEPARATOR_S,
+                                                   tpm_state_path, fn, NULL);
+            if (unlink(fullname) != 0) {
+                logerr(gl_LOGFILE, "Coud not remove %s: %s\n", fn, strerror(errno));
```
stefanberger, line 87: Coud -> Could

stefanberger (changes requested):

src/swtpm_setup/swtpm.c
PiMaker: it might actually be smarter to just do away with `swtpm_remove_pidfile` and unlink it immediately, that way no leftovers are possible - afaict this isn't used anywhere else, and I don't think there's much user value in keeping a pidfile for such a short-lived instance visible

src/swtpm_setup/swtpm_backend_dir.c
PiMaker: ack, missed it

stefanberger:
You probably saw Travis output. just missing header files...
```
swtpm_backend_dir.c: In function ‘parse_dir_state’:

swtpm_backend_dir.c:23:9: error: implicit declaration of function ‘strncmp’ [-Werror=implicit-function-declaration]

     if (strncmp(uri, "dir://", 6) == 0) {

         ^

swtpm_backend_dir.c: In function ‘delete_statefiles’:

swtpm_backend_dir.c:73:17: error: ‘errno’ undeclared (first use in this function)

             if (errno != 0 && errno != ENOENT) {

                 ^

swtpm_backend_dir.c:73:17: note: each undeclared identifier is reported only once for each function it appears in

swtpm_backend_dir.c:73:40: error: ‘ENOENT’ undeclared (first use in this function)

             if (errno != 0 && errno != ENOENT) {

                                        ^

swtpm_backend_dir.c:74:73: error: implicit declaration of function ‘strerror’ [-Werror=implicit-function-declaration]

             logerr(gl_LOGFILE, "Error getting next filename: %s\n", strerror(er

                                                                     ^
```

PiMaker:
no change, just rebased so CI can run again - the new test case should now pass

stefanberger:
> Oddly I get a test failure here on `tests/test_swtpm_setup_file_backend.log`:
My bad. Was working with old tree. Fixed now.

stefanberger:
Thanks!
Have a look at PR #576 .

Issue #559, opened 2021-09-26
Unable to build on ubuntu 21.04


odror:
**Describe the bug**
When building under Ubuntu 21.04 using the command "dpkg-buildpackage -us -uc -j32"
I get a fatal error
```
make[5]: *** [Makefile:731: test-suite.log] Error 1
make[5]: Leaving directory '/home/dror/work/swtpm/swtpm/tests'
make[4]: *** [Makefile:839: check-TESTS] Error 2
make[4]: Leaving directory '/home/dror/work/swtpm/swtpm/tests'
make[3]: *** [Makefile:1354: check-am] Error 2
make[3]: Leaving directory '/home/dror/work/swtpm/swtpm/tests'
make[2]: *** [Makefile:479: check-recursive] Error 1
make[2]: Leaving directory '/home/dror/work/swtpm/swtpm'
make[1]: *** [debian/rules:11: override_dh_auto_test] Error 2
make[1]: Leaving directory '/home/dror/work/swtpm/swtpm'
make: *** [debian/rules:4: binary] Error 2
dpkg-buildpackage: error: debian/rules binary subprocess returned exit status 2

```

**To Reproduce**
1. git clone https://github.com/stefanberger/swtpm
2. cd swtpm
3. dpkg-buildpackage -us -uc -j32
4.

**Expected behavior**
I expected to have a debian *.deb package

**Desktop (please complete the following information):**
 - OS: Ubuntu
 - Version 21.04

**Versions of relevant components**
- swtpm: 0.7.0~dev1.dsc
- libtpms: Not installed and not available for ubuntu
- openssl: 1.1.1j-1ubuntu3.5
- gnutls: Not installed and not available for ubuntu
- ...:

**Additional context**
Attached tests/test-suite.log
[test-suite.log](https://github.com/stefanberger/swtpm/files/7232232/test-suite.log)
.

stefanberger:
You have to first build libtpms using the instructions here: https://github.com/stefanberger/libtpms/wiki#build-a-package-on-ubuntu

The follow the instructions for building swtpm for Ubuntu here: https://github.com/stefanberger/swtpm/wiki#build-deb-package-ubuntu-debian

odror:
I have done it again remove and following  https://github.com/stefanberger/libtpms/wiki#build-a-package-on-ubuntu  
I get the same error
This is the procedure that I followed:

```
[...]

Issue #549, opened 2021-09-21
Cannot install swtpm on Ubuntu 20.04


elliotmtx:
OS: Ubuntu 20.04
I am trying to install swtpm and first I an below prereq step:

sudo apt-get -y install  libfuse-dev libglib2.0-dev libgmp-dev expect libtasn1-dev socat tpm-tools python3-twisted gnutls-dev gnutls-bin  libjson-glib-dev python3-setuptools softhsm2 libseccomp-dev
dpkg-buildpackage -us -uc -j$(nproc)

But it is failing with below error, kindly help

`Job for trousers.service failed because the control process exited with error code.
See "systemctl status trousers.service" and "journalctl -xe" for details.
invoke-rc.d: initscript trousers, action "start" failed.
● trousers.service - LSB: starts tcsd
   Loaded: loaded (/etc/init.d/trousers; generated; vendor preset: enabled)
   Active: failed (Result: exit-code) since Sun 2017-07-09 12:51:11 EDT; 5ms ago
     Docs: man:systemd-sysv-generator(8)
  Process: 9449 ExecStart=/etc/init.d/trousers start (code=exited, status=30)

Jul 09 12:51:11 s3id0n systemd[1]: Starting LSB: starts tcsd...
Jul 09 12:51:11 s3id0n trousers[9449]:  * Starting Trusted Computing daemon tcsd
Jul 09 12:51:11 s3id0n tcsd[9454]: TCSD TDDL[9454]: TrouSerS ioctl: (25) In…vice
Jul 09 12:51:11 s3id0n tcsd[9454]: TCSD TDDL[9454]: TrouSerS Falling back t…ort.
Jul 09 12:51:11 s3id0n trousers[9449]:    ...fail!
Jul 09 12:51:11 s3id0n systemd[1]: trousers.service: Control process exited…s=30
Jul 09 12:51:11 s3id0n systemd[1]: Failed to start LSB: starts tcsd.
Jul 09 12:51:11 s3id0n systemd[1]: trousers.service: Unit entered failed state.
Jul 09 12:51:11 s3id0n systemd[1]: trousers.service: Failed with result 'ex…de'.
Hint: Some lines were ellipsized, use -l to show in full.
dpkg: error processing package trousers (--configure):
 subprocess installed post-installation script returned error exit status 1
dpkg: dependency problems prevent configuration of tpm-tools:
 tpm-tools depends on trousers; however:
  Package trousers is not configured yet.

[...]

Issue #484, opened 2021-07-13
swtpm commit 5e73e32 causes apparmor failures on Ubuntu 20.04 hosts


galmasi:
**Describe the bug**

swtpm installed from source on Ubuntu 20.04 host, then running a TPM-enabled guest:
* the guest sees a TPM failure during boot
* the host sees apparmor failure messages related to swtpm during the same time

**To Reproduce**
1. Ubuntu 20.04 host (kernel 5.4.0-48-generic, libvirt 6.0.0-0ubuntu8.8) 
2. swtpm installed from source (commit 5e73e32) as per instructions: dependencies installed, libtpms installed from source.
3. Create any tpm2 enabled guest on libvirt.

**Expected behavior**

Guest boots up and `dmesg` reports a functional TPM2 device 

**Actual behavior**

Guest reports TPM self test failure:

```
[    1.392901] tpm_tis MSFT0101:00: 2.0 TPM (device-id 0x1, rev-id 1)
[    1.395489] tpm tpm0: A TPM error (257) occurred attempting the self test00001 BXPC 00000001)
[    1.398182] tpm tpm0: A TPM error (257) occurred attempting the self test
```

Host reports apparmor failure:

```
[25066717.437745] audit: type=1400 audit(1626115253.463:1314): apparmor="DENIED" operation="open" profile="libvirt-7cf29f14-a170-4e33-bece-1f1e4c7e10b7" name="/var/lib/libvirt/swtpm/7cf29f14-a170-4e33-bece-1f1e4c7e10b7/tpm2/" pid=2225183 comm="swtpm" requested_mask="r" denied_mask="r" fsuid=106 ouid=106
```

swtpm log shows file save failure:

```
SWTPM_NVRAM_StoreData: Error (fatal) opening /var/lib/libvirt/swtpm/7cf29f14-a170-4e33-bece-1f1e4c7e10b7/tpm2 for
 fsync failed, Permission denied
libtpms/tpm2: Entering failure mode; code: 3, location: ExecuteCommand line 308
libtpms/tpm2: TPM2_Process: Entered failure mode through command:
80 01 00 00 00 0c 00 00 01 44 00 00 
```

**Desktop (please complete the following information):**
 - OS: Ubuntu
 - Version 20.04

**Versions of relevant components**
- swtpm: 5e73e32
- libtpms: 98b3414
- openssl: 1.1.1f-1ubuntu2.3
- gnutls: 3.6.13-2ubuntu1.3
- ...:

[...]

PR #573, opened 2021-10-01
swtpm: Make fsync() errors non-fatal (for libvirt using AppArmor)

stefanberger:
Only recent libvirt versions have the patch for the AppArmor profile
for libvirt to allow fsync on dir and directory. Rather than failing
hard on this error, log it once and continue and do not try fsync
again after.

This patch addresses the problems seen on Ubuntu related to an older
version of libvirt without the AppArmor profile update.
- issue #484
- issue #549
- issue #559


stefanberger:
@Etsukata Any comments on this work-around?

Etsukata:
Thanks, this work-around (https://github.com/stefanberger/swtpm/pull/574)  looks good to me.

PR #576, opened 2021-10-05
Fix a couple of issue appearing on non-Linux systems

stefanberger:
This PR fixes a couple of issues on non-Linux systems.


PiMaker:
Can't comment on OSX compat, but the rest looks good to me.

PR #582, opened 2021-10-07
Debian improve postinst

ThomasLamprecht:
The first patch shouldn't raise too much objection, but I'm not 100% about the second one, why not create that directory on the fly when required or let all that handle tpm-udev[0], as that'd be required anyway if one wants to use the functionality where this directory is  actually needed, at least fwict.

[0]: https://salsa.debian.org/debian/tpm-udev/-/blob/master/debian/tpm-udev.postinst


stefanberger:
This is great. I had just thought about doing something like this this morning. So thanks a lot!

ThomasLamprecht:
hmm, weird I force pushed the branch to my fork 40 minutes ago, but it does not show up here. Not really used to the GH workflow, did I missed doing something?

PR #603, opened 2021-10-18
Prepare for v0.7.0 release

stefanberger:


elmarco:
Are you considering post-poning #461 to 0.8? I think that would make sense.

PR #619, opened 2021-10-29
Add support for --reconfigure for swtpm_setup to reconfigure the active PCR banks

stefanberger:
Add a --reconfigure option to swtpm_setup so that an existing state can be reconfigured
with a new set of active PCR banks.


elmarco:
lgtm

stefanberger:
Will apply once at least the discussions on libvirt mailing list have gone through.

PR #620, opened 2021-10-29
Openssl not certtool

vorlonofportland:
openssl (libcrypto) is a required dependency of swtpm_setup, but libggnutls is an optional dependency only required for swtpm_cert.  The openssl command is capable of doing everything certtool does, so use openssl to reduce dependencies.
    
This also fixes a documented bug that passwords must be passed on the commandline, because openssl supports passing them in by other methods.

The impetus for proposing this change is that certtool is in Ubuntu universe, but openssl is in Ubuntu main, so we would like to not have this as a runtime dependency.  (Keeping it as a build-time dependency for the tests is not a concern.)

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -294,23 +294,9 @@ if test "x$with_gnutls" != "xno"; then
     fi
 fi
 
-if test "x$with_gnutls" != "xno"; then
-    AC_PATH_PROG([GNUTLS_CERTTOOL], certtool)
-    if test "x$GNUTLS_CERTTOOL" = "x"; then
-        if test "x$with_gnutls" = "xyes"; then
-            AC_MSG_ERROR("Could not find certtool. Is gnutls-utils/gnutls-bin installed?")
-        else
-            with_gnutls=no
-        fi
-    fi
-    dnl certtool changed how it takes private key passwords
-    dnl 3.3.29 is too old (RHEL 7); we need at least gnutls 3.4.0
-    AC_MSG_CHECKING([for gnutls 3.4.0 or later])
-    $(pkg-config gnutls --atleast-version=3.4.0)
-    if test $? -ne 0; then
-        AC_MSG_ERROR([gnutls 3.4.0 is required])
-    fi
-    AC_MSG_RESULT([yes])
+AC_PATH_PROG([OPENSSL], openssl)
+if test "x$OPENSSL" = "x"; then
+    AC_MSG_ERROR("Could not find openssl. Is it installed?")
 fi
 
 if test "x$with_gnutls" != "xno"; then
--- a/configure.ac
+++ b/configure.ac
@@ -294,6 +294,25 @@ if test "x$with_gnutls" != "xno"; then
     fi
 fi
 
+if test "x$with_gnutls" != "xno"; then
+    AC_PATH_PROG([GNUTLS_CERTTOOL], certtool)
+    if test "x$GNUTLS_CERTTOOL" = "x"; then
+        if test "x$with_gnutls" = "xyes"; then
+            AC_MSG_ERROR("Could not find certtool. Is gnutls-utils/gnutls-bin installed?")
+        else
+            with_gnutls=no
+        fi
+    fi
+    dnl certtool changed how it takes private key passwords
+    dnl 3.3.29 is too old (RHEL 7); we need at least gnutls 3.4.0
+    AC_MSG_CHECKING([for gnutls 3.4.0 or later])
+    $(pkg-config gnutls --atleast-version=3.4.0)
+    if test $? -ne 0; then
+        AC_MSG_ERROR([gnutls 3.4.0 is required])
+    fi
+    AC_MSG_RESULT([yes])
+fi
+
 AC_PATH_PROG([OPENSSL], openssl)
 if test "x$OPENSSL" = "x"; then
     AC_MSG_ERROR("Could not find openssl. Is it installed?")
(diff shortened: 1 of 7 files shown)


stefanberger:
You will also have to make sure that all pkcs11 tests are working.

configure.ac
```diff
@@ -294,23 +294,9 @@ if test "x$with_gnutls" != "xno"; then
...
-        fi
-    fi
-    dnl certtool changed how it takes private key passwords
-    dnl 3.3.29 is too old (RHEL 7); we need at least gnutls 3.4.0
-    AC_MSG_CHECKING([for gnutls 3.4.0 or later])
-    $(pkg-config gnutls --atleast-version=3.4.0)
-    if test $? -ne 0; then
-        AC_MSG_ERROR([gnutls 3.4.0 is required])
-    fi
-    AC_MSG_RESULT([yes])
```
stefanberger, line 313: It's also in many test cases:
```
tests/common:   CERTTOOL=gnutls-certtool;;
tests/common:   CERTTOOL=certtool;;
```

vorlonofportland: Ok, do you prefer to keep the configure check in place for the test cases?

stefanberger: Unless we want to get rid of certtool also for the test cases I think we have no other choice. You could change the above now to state 'gnutls 3.4.0 is required for test cases'.

vorlonofportland:
> You will also have to make sure that all pkcs11 tests are working.

Which tests do you mean, specifically?  Are you seeing regressed tests with this change?  There are no regressions on Ubuntu with this change, but some of the tests are skipped here for various reasons including some that appear to encode Red Hat-specific paths (/usr/lib64/pkcs11/libtpm2_pkcs11.so).

stefanberger:
> 
> 
> > You will also have to make sure that all pkcs11 tests are working.
> 
> Which tests do you mean, specifically? Are you seeing regressed tests with this change? There are no regressions on Ubuntu with this change, but some of the tests are skipped here for various reasons including some that appear to encode Red Hat-specific paths (/usr/lib64/pkcs11/libtpm2_pkcs11.so).

My mistake. There shouldn't be any pkcs11 issues.

src/swtpm_localca/swtpm_localca.c
```diff
@@ -122,127 +119,200 @@ static int create_localca_cert(const gchar *lockfile, const gchar *statedir,
...
         cmd = concat_arrays(NULL,
                             (gchar *[]) {
-                                certtool,
-                                "--generate-self-signed",
-                                "--template", template1_file,
-                                "--outfile", cacert,
-                                "--load-privkey", cakey,
+                                openssl, "req", "-batch", "-x509",
+                                "-sha256", "-new",
+                                "-days", "10956",
```
stefanberger, line 212: It has to be a non-expiring certificate.

stefanberger: With certtool one could do that with -1 days...
```
        Validity
            Not Before: Nov  6 19:29:41 2021 GMT
            Not After : Dec 31 23:59:59 9999 GMT
```

src/swtpm_localca/swtpm_localca.c
```diff
@@ -122,127 +119,200 @@ static int create_localca_cert(const gchar *lockfile, const gchar *statedir,
...
                             (gchar *[]) {
-                                certtool,
-                                "--generate-certificate",
-                                "--template", template2_file,
-                                "--outfile", (gchar *)issuercert,
-                                "--load-privkey", (gchar *)signkey,
-                                "--load-ca-privkey", cakey,
-                                "--load-ca-certificate", cacert,
+                                openssl, "req", "-batch", "-sha256", "-new",
+                                "-days", "10956",
```
stefanberger, line 267: non-expiring also here

stefanberger (changes requested):

stefanberger:
> The impetus for proposing this change is that certtool is in Ubuntu universe, but openssl is in Ubuntu main, so we would like to not have this as a runtime dependency. (Keeping it as a build-time dependency for the tests is not a concern.)

And the next step is to modify swtpm_cert?

src/swtpm_localca/swtpm_localca.c
```diff
@@ -39,11 +39,7 @@ gchar *gl_LOGFILE = NULL;
 #define LOCALCA_OPTIONS "swtpm-localca.options"
 #define LOCALCA_CONFIG  "swtpm-localca.conf"
 
-#if defined __APPLE__
-# define CERTTOOL_NAME "gnutls-certtool"
-#else
-# define CERTTOOL_NAME "certtool"
-#endif
+#define CERTTOOL_NAME "openssl"
```
stefanberger, line 42: I think we can get rid of this now.

stefanberger:
@vorlonofportland ping ...

src/swtpm_localca/swtpm_localca.c
```diff
@@ -122,127 +119,200 @@ static int create_localca_cert(const gchar *lockfile, const gchar *statedir,
...
         /* create the root-CA's cert */
-        filecontent = "cn=swtpm-localca-rootca\n"
-                      "ca\n"
-                      "cert_signing_key\n"
-                      "expiration_days = -1\n";
-        template1_file_fd = write_to_tempfile(&template1_file,
-                                              (const unsigned char *)filecontent, strlen(filecontent));
-        if (template1_file_fd < 0)
+        filecontent = "HOME               = .\n"
+                      "RANDFILE           = $ENV::HOME/.rnd\n"
```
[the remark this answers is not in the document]
cpaelzer: FYI I also happened to find that this line is not evaluated correctly when calling openssl (the environment seems stripped at the time). So $HOME isn't set and it becomes $CWD/.rnd failing in plenty of cases. (see https://bugs.launchpad.net/ubuntu/+source/virt-manager/+bug/1968131)

@vorlonofportland are you in general going to overhaul this PR sooner or later?

stefanberger:
I am closing this PR also due lack of activity.

PR #630, opened 2021-11-08
build-sys: Move regex to test PCR banks into configure script

stefanberger:
Move the regex test for the PCR banks into the configure script.


stefanberger:
@elmarco This should resolve the issue forcing one to edit configure.ac to change the PCR bank list.

elmarco:
ok, lgtm

stefanberger:
@elmarco Thanks.

PR #636, opened 2021-11-22
Move *.conf and *.options to man5

stefanberger:
According to the man page sections guideline, man8 should be used
for system administration commands. So this commit moves *.conf and
*.options files to man5.

Signed-off-by: Seunghun Han <kkamagui@gmail.com>


stefanberger:
@kkamagui I added swtpm.spec to it.

Issue #671, opened 2022-03-03
swtpm --daemon fails with relative path for --tpmstate


osresearch:
**Describe the bug**
Invoking `swtpm --daemon` with a relative path for `--tpmstate dir=./foo` exits successfuly, but when `tpm2-tools` tries to connect to it produces a "`commands not being accepted because of a TPM failure`" error.

**To Reproduce**
This works:
```
swtpm socket \
  --daemon \
  --tpmstate dir=/tmp \
  --server type=unixio,path=/tmp/swtpm-sock \
  --ctrl type=unixio,path=/tmp/tpm-sock.ctrl \
  --flags not-need-init,startup-clear \
  --tpm2
TPM2TOOLS_TCTI=swtpm:path=/tmp/swtpm-sock tpm2 pcrread sha256:0
  sha256:
    0 : 0x0000000000000000000000000000000000000000000000000000000000000000
```
but invoking it with a relative path for the `--tpmstate dir` produces an odd error when invoked with `--daemon`:
```
# mkdir ./tpm-state
# swtpm socket \
  --daemon \
  --tpmstate dir=./tpm-state \
  --server type=unixio,path=/tmp/swtpm-sock \
  --ctrl type=unixio,path=/tmp/tpm-sock.ctrl \
  --flags not-need-init,startup-clear \
  --tpm2
# echo $?
0
# TPM2TOOLS_TCTI=swtpm:path=/tmp/swtpm-sock tpm2 pcrread sha256:0
WARNING:esys:src/tss2-esys/api/Esys_GetCapability.c:303:Esys_GetCapability_Finish() Received TPM Error 
ERROR:esys:src/tss2-esys/api/Esys_GetCapability.c:107:Esys_GetCapability() Esys Finish ErrorCode (0x00000101) 
ERROR: Esys_GetCapability(0x101) - tpm:error(2.0): commands not being accepted because of a TPM failure
ERROR: Unable to run pcrread
```

Running it with a relative path and putting it into the background with the shell also works.
```
mkdir ./tpm-state
swtpm socket \
  --tpmstate dir=./tpm-state \
  --server type=unixio,path=/tmp/swtpm-sock \
  --ctrl type=unixio,path=/tmp/tpm-sock.ctrl \
  --flags not-need-init,startup-clear \
  --tpm2 &
[1] 2057740
TPM2TOOLS_TCTI=swtpm:path=./tpm-state/swtpm-sock tpm2 pcrread sha256:0
  sha256:
[...]

PR #672, opened 2022-03-03
swtpm: Do not chdir(/) when using --daemon

stefanberger:
With relative paths being used the chdir("/") in daemonize_finish() will
cause file access errors.

Fixes: 98d1d12 ("swtpm: Make --daemon not racy")
Resolves: https://github.com/stefanberger/swtpm/issues/671

### DIFF
--- a/src/swtpm/daemonize.c
+++ b/src/swtpm/daemonize.c
@@ -274,12 +274,6 @@ daemonize_finish(void)
         return;
     }
 
-    if (chdir("/") == -1) {
-        fprintf(stderr, "Failed to change directory to /: %s\n",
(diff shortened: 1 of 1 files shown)


stefanberger:
@nicowilliams Please have a look.

nicowilliams:
Ah, of course.  Sorry, I was just doing what is _typically_ expected of daemonization, but that `chdir()` was not needed here.

stefanberger:
Actually, it had been a long-standing bug still from calling `daemon()` before.

nicowilliams:
In that case, I helped fix it :)

PR #701, opened 2022-05-30
Send a TPM2_Shutdown before TPMLIB_Terminate() in case client did not send it

stefanberger:
If necessary send a TPM2_Shutdown() command to libtpms before processing
CMD_INIT. However, this is only necessary for a TPM 2 and only if the
TPM2_Shutdown command has not been sent by the client (VM TPM driver) as
the last command as it should do under normal circumstances, for example
upon graceful VM shutdown.

This fixes a bug where abrupt VM resets may trigger the TPM 2's dictionary
attack lockout logic due to the TPM 2 not having received a TPM2_Shutdown
command before it was reset using CMD_INIT for example. An OS driver is
typically supposed to send a TPM2_Shutdown to the TPM 2 but an abrupt VM
reset prevents it.

There are 3 control commands where this needs to be done since they
call TPMLIB_Terminate():

- CMD_STOP:
   This command is typically called before setting the state blobs of the
   TPM or before configuring the buffer size [QEMU, test cases].

- CMD_INIT:
   This command is called for resetting and initializing the TPM 2.

- CMD_SHUTDOWN:
   This command is called for a graceful shutdown of the TPM 2.

There are no negative side effects to be expected if TPM2_Shutdown()
is sent before any of these. Also, since none of these are sent before
the state of the TPM is marshalled (for migration for example) migrated
state will not have a TPM2_Shutdown() applied to it (accidentally).

Edk2 sends a sequence of TPM2_Shutdown(SU_STATE) + TPM2_GetRandom()
before suspend-to-ram. Upon wake up a CMD_INIT is sent to the TPM to
reset it, which in this case now requires a TPM2_Shutdown(SU_STATE)
to be sent to the TPM 2 so that certain TPM 2 state is available
again upon resume. To avoid invaliding the SU_STATE, first send a
TPM2_Shutdown(SU_STATE) in *all cases* and only if this fails send a
TPM2_Shutdown(SU_CLEAR). This way the internal state is preserved and
the VM (or user) are expected to use TPM2_Startup(SU_CLEAR) when
staring up the TPM 2 and no previous state needs to be resumed.

Note: The TPM 2 spec describes the command as follows:

"This command is used to prepare the TPM for a power cycle. The
shutdownType parameter indicates how the subsequent TPM2_Startup() will be
processed.[...]
This command saves TPM state but does not change the state other than the
internal indication that the context has been saved. The TPM shall
continue to accept commands. If a subsequent command changes TPM state
saved by this command, then the effect of this command is nullified. The
TPM MAY nullify this command for any subsequent command rather than check
whether the command changed state saved by this command. If this command
is nullified and if no TPM2_Shutdown() occurs before the next
TPM2_Startup(), then the next TPM2_Startup() shall be
TPM2_Startup(CLEAR)."

Buglink: https://bugzilla.redhat.com/show_bug.cgi?id=2087538

### DIFF
--- a/man/man8/swtpm.pod
+++ b/man/man8/swtpm.pod
@@ -406,6 +406,19 @@ Display usage info.
 
 =back
 
+=head1 NOTES
+
+If a TPM 2 is used, the user is typically required to send a TPM2_Shutdown()
+command to a TPM 2 to avoid possibly increasing the TPM_PT_LOCKOUT_COUNTER
+that may lead to a dictionary attack (DA) lockout upon next startup
+(TPM2_Startup()) of the TPM 2. Whether the TPM_PT_LOCKOUT_COUNTER is
+increased depends on previous commands sent to the TPM 2 as well as
+internal state of the TPM 2. One example that will trigger the counter to
+increase is the omission of a password when trying to access a
+password-protected object or NVRAM location that has the DA attribute set,
+followed by termination of swtpm without sending TPM2_Shutdown(). To avoid
+a DA lockout swtpm will make a best-effort and send a TPM2_Shutdown(SU_STATE)
+or TPM2_Shutdown(SU_CLEAR) if found necessary.
 
 =head1 SEE ALSO
 
--- a/man/man8/swtpm.pod
+++ b/man/man8/swtpm.pod
@@ -277,7 +277,7 @@ The I<log> action is only available if libseccomp supports logging.
 This option is only available on Linux and only if swtpm was compiled with
 libseccomp support.
 
-=item B<--flags [not-need-init] [,startup-clear|startup-state|startup-deactivated|startup-none]>
+=item B<--flags [not-need-init][,startup-clear|startup-state|startup-deactivated|startup-none][,disable-auto-shutdown]>
 
 The I<not-need-init> flag enables the TPM to accept TPM commands right after
 start without requiring an INIT to be sent to it through the command channel
@@ -290,6 +290,13 @@ I<startup-none> option, which results in no command being sent.
 
 If I<--vtpm-proxy> is used, I<startup-clear> is automatically chosen but
 this can be changed with this option.
+
+The I<disable-auto-shutdown> flag prevents swtpm from automatically sending a
+TPM2_Shutdown() before the reset of a TPM 2 or before the swtpm process
+is terminated. When this flag is not provide swtpm will send this command to
+avoid increasing the dictionary attack (DA) lockout counter and ulimately
+a DA lockout by the TPM 2 due to omission of sending a required TPM2_Shutdown()
+before TPM 2 reset or swtpm process termination.
 
 =item B<--print-capabilities> (since v0.2)
 
@@ -309,6 +316,7 @@ may contain the following:
         "nvram-backend-file",
         "tpm-send-command-header",
         "flags-opt-startup",
+        "flags-opt-disable-auto-shutdown",
         "rsa-keysize-1024",
         "rsa-keysize-2048",
         "rsa-keysize-3072"
@@ -368,6 +376,10 @@ The TPM 2 will respond by preprending a 4-byte response indicator and a
 =item B<flags-opt-startup> (since v0.3)
 
 The I<--flags> option supports the I<startup-...> options.
+
+=item B<flags-opt-disable-auto-shutdown> (since v0.8)
+
+The I<--flags> option supports the I<disable-auto-shutdown> flag.
 
 =item B<rsa-keysize-2048> (since v0.4)
 
--- a/src/swtpm/mainloop.c
+++ b/src/swtpm/mainloop.c
@@ -102,6 +102,7 @@ int mainLoop(struct mainLoopParams *mlp,
     struct iovec        iov[3];
     uint32_t            ack = htobe32(0);
     struct tpm2_resp_prefix respprefix;
+    uint32_t            lastCommand;
 
     /* poolfd[] indexes */
     enum {
@@ -141,9 +142,11 @@ int mainLoop(struct mainLoopParams *mlp,
                                   mlp->startupType,
                                   mlp->tpmversion,
                                   command, max_command_length);
-        if (command_length > 0)
+        if (command_length > 0) {
+            mlp->lastCommand = tpmlib_get_cmd_ordinal(command, command_length);
             rc = TPMLIB_Process(&rbuffer, &rlength, &rTotal,
                                 command, command_length);
+        }
 
         if (rc || command_length == 0) {
             mainloop_terminate = true;
@@ -271,6 +274,14 @@ int mainLoop(struct mainLoopParams *mlp,
                                                       mlp->tpmversion);
                     goto skip_process;
                 }
+            }
+
+            if (rc == 0) {
+                lastCommand =
+                    tpmlib_get_cmd_ordinal(&command[cmd_offset],
+                                           command_length - cmd_offset);
+                if (lastCommand != TPM_ORDINAL_NONE)
+                    mlp->lastCommand = lastCommand;
             }
 
             if (rc == 0) {
--- a/src/swtpm/mainloop.h
+++ b/src/swtpm/mainloop.h
@@ -60,6 +60,7 @@ struct mainLoopParams {
     uint32_t locality_flags;
     TPMLIB_TPMVersion tpmversion;
     uint16_t startupType; /* use TPM 1.2 types */
+    uint32_t lastCommand; /* last Command sent to TPM */
 };
 
 int mainLoop(struct mainLoopParams *mlp,
--- a/src/swtpm/mainloop.h
+++ b/src/swtpm/mainloop.h
@@ -61,6 +61,7 @@ struct mainLoopParams {
     TPMLIB_TPMVersion tpmversion;
     uint16_t startupType; /* use TPM 1.2 types */
     uint32_t lastCommand; /* last Command sent to TPM */
+    bool disable_auto_shutdown;   /* TPM2_Shutdown() will NOT be sent by swtpm */
 };
 
 int mainLoop(struct mainLoopParams *mlp,
--- a/src/swtpm/capabilities.c
+++ b/src/swtpm/capabilities.c
@@ -146,7 +146,7 @@ int capabilities_print_json(bool cusetpm, TPMLIB_TPMVersion tpmversion)
          "{ "
          "\"type\": \"swtpm\", "
          "\"features\": [ "
-             "%s%s%s%s%s%s%s%s%s%s%s"
+             "%s%s%s%s%s%s%s%s%s%s%s%s"
           " ], "
          "\"version\": \"" VERSION "\" "
          "}",
@@ -154,6 +154,7 @@ int capabilities_print_json(bool cusetpm, TPMLIB_TPMVersion tpmversion)
          with_tpm2,
          !cusetpm     ? "\"tpm-send-command-header\", ": "",
          true         ? "\"flags-opt-startup\", "      : "",
+         true         ? "\"flags-opt-disable-auto-shutdown\", ": "",
          cmdarg_seccomp,
          true         ? "\"cmdarg-key-fd\", "          : "",
          true         ? "\"cmdarg-pwd-fd\", "          : "",
--- a/src/swtpm/common.c
+++ b/src/swtpm/common.c
@@ -246,6 +246,9 @@ static const OptionDesc flags_opt_desc[] = {
     }, {
         .name = "startup-deactivated",
         .type = OPT_TYPE_BOOLEAN,
+    }, {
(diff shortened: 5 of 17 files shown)


stefanberger:
I will remove the warning print-out. It will become a hidden feature of swtpm then to send TPM2_Shutdown when needed.

```
$ ./test_tpm2_probe
Need to be root to run test with CUSE interface.
Need to be root to run test with CUSE interface.
==== Starting swtpm with interfaces socket+socket ====
swtpm: Warning: Need to send TPM2_Shutdown(SU_CLEAR); previous command: 0x0000017e
OK
==== Starting swtpm with interfaces socket+unix ====
swtpm: Warning: Need to send TPM2_Shutdown(SU_CLEAR); previous command: 0x0000017e
OK
==== Starting swtpm with interfaces unix+unix ====
swtpm: Warning: Need to send TPM2_Shutdown(SU_CLEAR); previous command: 0x0000017e
OK
```

src/swtpm/mainloop.c
```diff
@@ -102,6 +102,7 @@ int mainLoop(struct mainLoopParams *mlp,
     struct iovec        iov[3];
     uint32_t            ack = htobe32(0);
     struct tpm2_resp_prefix respprefix;
+    uint32_t            lastCommand;
```
elmarco, line 105: is this state being migrated?

elmarco:
looks good otherwise

src/swtpm/mainloop.c
stefanberger: No.

stefanberger: Sending a TPM2_Shutdown() twice before a TPMLIB_Terminate(), will not affect the state of the TPM 2. Actually the spec says:

https://trustedcomputinggroup.org/wp-content/uploads/TCG_TPM2_r1p64_Part3_Commands_15may2021.pdf

"This command is used to prepare the TPM for a power cycle. The shutdownType parameter indicates
how the subsequent TPM2_Startup() will be processed.

[...]
This command saves TPM state but does not change the state other than the internal indication that the
context has been saved. The TPM shall continue to accept commands. If a subsequent command
changes TPM state saved by this command, then the effect of this command is nullified. The TPM MAY
nullify this command for any subsequent command rather than check whether the command changed
state saved by this command. If this command is nullified. and if no TPM2_Shutdown() occurs before the
next TPM2_Startup(), then the next TPM2_Startup() shall be TPM2_Startup(CLEAR)."

stefanberger:
> Looks good otherwise. It would be nice to have a test for it, and some quote of the specification

It will  become silent. And for RHEL I would recommend taking the latest version 0.7.3 + these patches of wait for 0.7.4, but this may take a bit...

stefanberger:
I have looked into how to trigger the DA counter to increase upon reset but couldn't figure it out so far...

src/swtpm/mainloop.c
stefanberger: > is this state being migrated?

Phew, actually that could in one cornercase become a problem. If the machine migrates with last command sent TPM2_Shutdown(SU_STATE)... swtpm writes not state at all today but only libtpms.

stefanberger: ... but then on the other hand the TPM2_Shutdown() that this patch here is sending only happens after the mirgation occurred, so it shouldn't affect it at all.

stefanberger: @elmarco VMs ins power management mode don't go down iirc, they just sit there and do nothing until one presses a key or so, right?

elmarco: Power management is a complex topic ;). But  VMs are generally simply suspended  or "paused" (not using PM).

The trouble is that if the VM is "live" migrated (saved & restored without PM suspend etc), it will lose the lastCommand state... We probably need another GET_STATEBLOB type for swtpm data.

stefanberger: > "Just how often is this going to happen?" live migration? it depends. save/restore are common operations. Some centers balance the load with live migration too.

Live migration itself is not the problem. Live migration and then abrupt termination of the VM would be the issue. Any normal OS is typically supposed to be shut down and the TPM 2 just happens to require that the OS be properly shut down.

elmarco: Well then, do we actually need to do anything? DA lockout might just be working as intended.

stefanberger: A one-line solution for this problem maybe here:

https://github.com/stefanberger/libtpms/blob/9165880bf72c26154b4b2aa354c89cb86a5e7a7a/src/tpm2/DA.c#L136-L150
```
    // If DA has not been disabled and the previous shutdown is not orderly
    // failedTries is not already at its maximum then increment 'failedTries'
    if(gp.recoveryTime != 0
       && gp.failedTries < gp.maxTries
       && !IS_ORDERLY(g_prevOrderlyState))
	{
#if USE_DA_USED
	    gp.failedTries += g_daUsed;
	    g_daUsed = FALSE;
#else
	    gp.failedTries++;
#endif
	    // Record the change to NV
	    NV_SYNC_PERSISTENT(failedTries);
	}
```

We could just comment the line `gp.failedTries += g_daUsed;`. That should have the same effect.

elmarco: > As for this series, it addresses the issue on the swtpm level for most scenarios except for migration and then abrupt termination case where no further command was sent to the TPM by the VM. I think in this case I would not go through the extra mile of migrating the lastCommand to the destination but let the TPM 2 specs kick in demanding proper shutdown.

If that's what the spec demands, then why do anything at all? let's not add different behaviours when the maching is migrated or not, please :)

stefanberger: I would say this PR should address 99% of the cases where an abrupt shutdown happens. If someone wants to trigger the exception then they migrate a VM multiple times and terminate it every time to run into that one case that is not covered.

stefanberger:
In the end there's one concern that I have about this series and that is that we accidentally send a TPM2_Shutdown(SU_CLEAR) after a TPM2_Shutdown(SU_STATE) was issued. The relevant scenarios to trigger this are `systemctl suspend` and `systemctl hibernate` that both send a TPM2_Shutdown(SU_STATE) but then also completely shut down the VM. The BIOS upon boot then uses TPM2_Startup(SU_CLEAR) clearing the state and the TPM2_Shutdown(SU_STATE) was useless. So I am wondering whether this is normal expected behavior for an x86 VM that upon `systemctl suspend` [and `systemctl hibernate`] the VM completely shuts down.

stefanberger:
I did some more tests with suspend-to-ram (systemctl suspend) with EDK2. Here the Linux drivers sends TPM2_Shutdown(SU_STATE) and then a TPM2_GetRandom() is sent to the TPM 2 before the VM suspends. Once the VM awakes again, QEMU resets the TIS or CRB interface and swtpm is also reset via CMD_INIT (as if the whole VM was reset). We cannot simply send a TPM2_Shutdown(SU_CLEAR) before the CMD_INIT since the TPM2_Shutdown(SU_STATE) was sent but the lastCommand would be TPM2_CC_GetRandom. So, determining whether to send a TPM2_Shutdown() by looking at the lastCommand that was sent to the TPM doesn't always work and in this case it would mess up the state of the TPM.

Note: EDK2's (of QEMU 7.0) S3 resume code seems broken since it didn't send a TPM2_Startup(SU_STATE) upon resume.

stefanberger:
> I did some more tests with suspend-to-ram (systemctl suspend) with EDK2. Here the Linux drivers sends TPM2_Shutdown(SU_STATE) and then a TPM2_GetRandom() is sent to the TPM 2 before the VM suspends. Once the VM awakes again, QEMU resets the TIS or CRB interface and swtpm is also reset via CMD_INIT (as if the whole VM was reset). We cannot simply send a TPM2_Shutdown(SU_CLEAR) before the CMD_INIT since the TPM2_Shutdown(SU_STATE) was sent but the lastCommand would be TPM2_CC_GetRandom. So, determining whether to send a TPM2_Shutdown() by looking at the lastCommand that was sent to the TPM doesn't always work and in this case it would mess up the state of the TPM.

The solution is to send a TPM2_Shutdown(SU_STATE) in case lastCommand is not TPM2_Shutdown. If this command fails TPM2_Shutdown(SU_CLEAR) is sent. The SU_STATE stores additional state that TPM2_Startup(SU_STATE) can then restore and TPM2_Startup(SU_CLEAR) would simply ignore. Now we have to trust the BIOS to use the right command and not cause stored state to be restored unnecessarily. During normal startup the BIOS would use SU_CLEAR, so we should be fine.

stefanberger:
Once  I have confirmation that this patch resolve the issue presented in the bugzilla I will merge it.

src/swtpm/mainloop.c
elmarco: We should leave a FIXME for migration support.

stefanberger:
I was planning to merge this later this week unless there are comments or concerns.

man/man8/swtpm.pod
```diff
@@ -406,6 +406,15 @@ Display usage info.
...
 
+=head1 NOTES
+
+If a TPM 2 is used, the user is typically required to send a TPM2_Shutdown()
+command to a TPM 2 to avoid possibly increasing the TPM_PT_LOCKOUT_COUNTER
+that may lead to a dictionary attack lockout upon next startup
+(TPM2_Startup()) of the TPM 2. Whether the TPM_PT_LOCKOUT_COUNTER is
+actually increased depends on previous commands sent to the TPM 2 as well as
+internal state of the TPM 2. Swtpm will make a best-effort and send a
+TPM2_Shutdown(SU_STATE) or TPM2_Shutdown(SU_CLEAR) if found necessary.
```
elmarco, line 417: Please be more specific on when this happen, otherwise a casual reader will wonder and need to look at the sources.

src/swtpm/mainloop.h
```diff
@@ -60,6 +60,7 @@ struct mainLoopParams {
     uint32_t locality_flags;
     TPMLIB_TPMVersion tpmversion;
     uint16_t startupType; /* use TPM 1.2 types */
+    uint32_t lastCommand; /* last Command sent to TPM */
```
elmarco, line 63: So this state is still not transferred during migration? I am afraid it is required for proper support.

src/swtpm/mainloop.c
elmarco: Migration doesn't imply suspend-to-ram.

src/swtpm/mainloop.h
stefanberger: Can you explain why because I don't 'see' it.

man/man8/swtpm.pod
stefanberger: I updated the text now with an example.

src/swtpm/mainloop.c
stefanberger: Correct, it doesn't. 

When doing suspend-to-ram with EDK2 the last-sent command is unfortunately not a TPM2_Shutdown() but a TPM2_GetRandom(). In fact EDK2 could legitimately send yet other commands than the TPM2_GetRandom() *after* the TPM2_Shutdown() without affecting the correct functioning of TPM2_Startup(SE_STATE) upon resume. So tracking the last-send command can only be used as an optimization.
Upon resume from suspend-to-ram swtpm needs to handle a CMD_INIT (due to reset) which looks like the abrupt reset case where we need to send a TPM2_Shutdown() [original purpose of this patch]. Now in this case with EDK2 (or any other firmware) we may not have seen TPM2_Shutdown(SU_STATE) as the last command but that TPM2_GetRandom() instead. To not mess up the TPM 2's state for proper resume with TPM2_Startup(SU_STATE), swtpm needs to send a TPM2_Shutdown(SU_STATE) to the TPM 2, which we obviously do for the original reason to avoid DA lockouts. So this is where the *SU_STATE* versus *SU_CLEAR* in the TPM2_Shutdown(SU_STATE) comes from.

stefanberger:
I am wondering whether to add a flag `--flags auto-shutdown` to enable this feature.

src/swtpm/mainloop.h
elmarco: If the machine shuts down before migration (or saved), reaches target and boot, you don't have the state saved.

Either you have a useful state that must be saved, or you don't and then you don't need it at all.

elmarco:
> I am wondering whether to add a flag `--flags auto-shutdown` to enable this feature.

Perhaps the other way around (disable-auto-shutdown), if we believe the new behavior is a fix and it doesn't break migration etc.

stefanberger:
> Perhaps the other way around (disable-auto-shutdown), if we believe the new behavior is a fix and it doesn't break migration etc.

I don't think it breaks anything. It rather fixes an unexpected behavior of the TPM 2 for the virtualization case though an abrupt reset on a physical machine may cause the same.
I'll do the disable-auto-shutdown instead now.

PR #704, opened 2022-06-08
swtpm: Disable OpenSSL FIPS mode to avoid libtpms failures

stefanberger:
While libtpms does not provide any means to disable FIPS-disabled crypto
algorithms from being used, work around the issue by simply disabling the
FIPS mode of OpenSSL if it is enabled. If it cannot be disabled, exit
swtpm with a failure message that it cannot be disabled. If FIPS mode
was successfully disabled, print out a message as well.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -156,6 +156,15 @@ openssl)
 	AC_MSG_RESULT([Building with openssl crypto library])
 	LIBCRYPTO_LIBS=$(pkg-config --libs libcrypto)
 	AC_SUBST([LIBCRYPTO_LIBS])
+	AC_CHECK_HEADERS([openssl/fips.h],
+	                 [AC_DEFINE_UNQUOTED([HAVE_OPENSSL_FIPS_H], 1,
+	                                     [whether openssl/fips.h is available])]
+	                 )
+	AC_CHECK_LIB(crypto,
+		     [FIPS_mode_set],
+		     [AC_DEFINE_UNQUOTED([HAVE_OPENSSL_FIPS_MODE_SET_API], 1,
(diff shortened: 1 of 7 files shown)


stefanberger:
@elmarco This is the solution for now to run swtpm on FIPS-enabled machines.

src/swtpm/utils.c
```diff
@@ -329,3 +334,43 @@ ssize_t read_eintr(int fd, void *buffer, size_t buflen)
...
+#ifdef HAVE_OPENSSL_FIPS_H
+int disable_fips_mode(void)
+{
+    int mode = FIPS_mode();
+    int ret = 0;
+
+    if (mode != 0) {
+#if OPENSSL_VERSION_NUMBER >= 0x30000000L
+        logprintf(STDERR_FILENO,
+                  "FIPS mode cannot be disabled with OpenSSL 3.x\n");
```
berrange, line 353: The openssl 3 'migration_guide' man page says

```
       Removed FIPS_mode() and FIPS_mode_set()

       These functions are legacy APIs that are not applicable to the new
       provider model. Applications should instead use
       EVP_default_properties_is_fips_enabled(3) and
       EVP_default_properties_enable_fips(3).
```
so we don't need to stub this out as an error AFAICT

stefanberger: I ended up using the recommented EVP APIs...

stefanberger:
> Obviously this change would prevent the vTPM actually being used with/included in any government certification programs for virt hosts that require FIPS, but as a short term workaround to at least get the VMs booting, the code looks OK.

So this patch works in the scenario described in the bugzilla?

PR #713, opened 2022-07-11
[DRAFT] SWTPM support Libressl on Linux

williamcroberts:


stefanberger:
I will probably mark it as experimental because I will likely not test much with now two different crypto libs.

williamcroberts:
This works AS IS as long as you use version libressl-v3.1.0 which includes the required RSA OAEP support.

PR #714, opened 2022-07-12
configure: check for bash

williamcroberts:
PCR Bank verification needs bash, so check for bash. While at it use the
autoconf shell construct macros over raw shell syntax which is slightly
more portable.

Examples:
./configure --enable-default-pcr-banks=sha256,sha920
checking which PCR banks to activate by default... configure: error: sha256,sha920 is an invalid list of PCR banks

./configure --enable-default-pcr-banks=sha256,sha512
checking which PCR banks to activate by default... sha256,sha512

./configure
checking which PCR banks to activate by default... sha256

Signed-off-by: William Roberts <william.c.roberts@intel.com>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -366,16 +366,22 @@ AC_ARG_ENABLE([default-pcr-banks],
               []
 )
 
-if test "x$enable_default_pcr_banks" != "x"; then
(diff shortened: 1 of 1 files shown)


stefanberger:
Thanks. LGTM.

PR #718, opened 2022-07-19
Add compatibility for RHEL 7 GnuTLS

rosslagerwall:
Tested with 3.3.26 and 3.6.16.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -312,14 +312,6 @@ if test "x$with_gnutls" != "xno"; then
             with_gnutls=no
         fi
     fi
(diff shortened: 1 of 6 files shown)


stefanberger:
What's the incentive to support RHEL 7?

stefanberger:
I previously had support for older versions of GnuTLS but dropped it since swtpm really is only being used with more recent versions of GnuTLS. So my preference is to leave the code as-is and only support newer versions of GnuTLS -- which will hopefully require no changes.

PR #732, opened 2022-08-11
Release the lock on the storage backend when migrating with shared storage

stefanberger:
This PR adds a `--migration` command line option with parameters `incoming` and `release-lock-outgoing` that allows to migrate the state over shared storage of the swtpm's state directory (between source and destination) by deferring the locking in incoming migration and releasing the storage once outgoing migration is retrieving the 'savestate' of the TPM. To accommodate a fallback in case of migration failure (QEMU migration failed) the re-locking of the storage is attempted several times until swtpm on the migration destination side has terminated and released the lock there.


stefanberger:
The patch to apply to this to **simulate** `--migration incoming,release-lock-outgoing` on the command line:

```
From 8241c2250eef90088e1126365953745bd919e395 Mon Sep 17 00:00:00 2001
From: Stefan Berger <stefanb@linux.ibm.com>
Date: Wed, 10 Aug 2022 19:35:41 -0400
Subject: [PATCH] temp: Simulate '--migration incoming,release-lock-outgoing'
 by setting flags

This breaks the test case in this series!
---
 src/swtpm/swtpm.c | 2 ++
 1 file changed, 2 insertions(+)

diff --git a/src/swtpm/swtpm.c b/src/swtpm/swtpm.c
index aa14066..47b11ee 100644
--- a/src/swtpm/swtpm.c
+++ b/src/swtpm/swtpm.c
@@ -486,6 +486,8 @@ int swtpm_main(int argc, char **argv, const char *prgname, const char *iface)
         goto exit_failure;
     }

+    mlp.incoming_migration = mlp.release_lock_outgoing = true;
+
     if (server) {
         if (server_get_fd(server) >= 0) {
             mlp.fd = server_set_fd(server, -1);
--
2.37.1

```

stefanberger:
This is a bit of a concern -- very difficult/almost impossible to test and trigger and verify in a real-world scenario:
```
When migrating a VM the migration may fail and execution will then resume
one the originating side. In this fallback case the swtpm on the
destination side may need some time to terminate and release the lock.
Therefore, add a loop to the code attempting to re-lock the storage
directory on the source side for a few times until on the destination
side swtpm has released the lock. Retry the locking for 100 times
with 10ms in between. The retries will only ever be necessary of a TPM 2
command is immediately executed upon resume and this may be difficult
to test. The negative side effects of this could be that the loop is not
long enough to grab the lock or that a short command sent by the TPM
driver will time out due to the retries.
```

man/man8/swtpm.pod
```diff
@@ -412,6 +417,23 @@ the 'permall' and 'volatile' states are available.
...
+
+This option allows to control the locking of the NVRAM storage
+for the purpose of supporting migration between hosts that have
+shared storage setup for the swtpm's state directory and if
+locking is supported by the storage backend. The directory storage
+backend for example supports locking and therefore requires usage
+of this option in case of shared storage.
+When providing the I<incoming> option parameter swtpm will defer the
+locking of the NVRAM until the state blobs are received or until the
+first TPM command is processed if no state blobs are received.
```
stefanberger, line 430: are -> were

src/swtpm/swtpm_chardev.c
```diff
@@ -207,6 +207,10 @@ static void usage(FILE *file, const char *prgname, const char *iface)
     "                 : Choose the action of the seccomp profile when a\n"
     "                   blacklisted syscall is executed; default is kill\n"
 #endif
+    "--migration [incoming][,release-lock-outgoing]\n"
+    "                 : Incoming migration will defer locking of NVRAM storage\n"
+    "                   Release-lock-outgoing will release the NVRAM lock on\n"
```
stefanberger, line 212: R->r

stefanberger:
libvirt support for shared storage is here: https://github.com/stefanberger/libvirt-tpm/tree/master%2Bswtpm_shared_storage.v1

I'll post these patches on the libvirt mailing list soon.

PR #740, opened 2022-08-23
swtpm: Add a chroot option

rosslagerwall:
Add an option to enter a chroot after starting swtpm. This is useful for
sandboxing purposes. When this option is used, it is expected that swtpm
is started as root and the --runas option is used to subsequently drop
privileges (otherwise the chroot could be escaped).

Signed-off-by: Jennifer Herbert <jennifer.herbert@citrix.com>
Signed-off-by: Ross Lagerwall <ross.lagerwall@citrix.com>

### DIFF
--- a/src/swtpm/swtpm.c
+++ b/src/swtpm/swtpm.c
@@ -175,6 +175,8 @@ static void usage(FILE *file, const char *prgname, const char *iface)
     "                   disable-auto-shutdown disables automatic sending of\n"
     "                   TPM2_Shutdown before TPM 2 reset or swtpm termination;\n"
     "-r|--runas <user>: change to the given user\n"
+    "-R|--chroot <path>\n"
+    "                 : chroot to the given directory at startup\n"
     "--tpm2           : choose TPM2 functionality\n"
 #ifdef WITH_SECCOMP
 # ifndef SCMP_ACT_LOG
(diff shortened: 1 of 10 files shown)


stefanberger:
Thanks. The patch looks good to me. Ideally there would be a test case for this as well, one derived from an existing test case should be possible to add.

src/swtpm/swtpm.c
```diff
@@ -249,6 +250,7 @@ int swtpm_main(int argc, char **argv, const char *prgname, const char *iface)
         {"fd"        , required_argument, 0, 'f'},
         {"server"    , required_argument, 0, 'c'},
         {"runas"     , required_argument, 0, 'r'},
+        {"chroot"    , required_argument, 0, 'R'},
```
stefanberger, line 253: You have to add this also to the help screens.

rosslagerwall: OK, will do. I'll also add a test case.

stefanberger: If you need a pair of available TCP ports for the test cases, please use 65468/65469

rosslagerwall: Updated as suggested.

gramine4aarch64:
Unfortunately I cannot merge it unless you rebase it on latest master...

stefanberger:
Merged now. thanks.

stefanberger:
Does this work with the CUSE interface for you?

stefanberger:
> Does this work with the CUSE interface for you?

It did not work for me since on my system tpmfs is mounted with the nodev option.

rosslagerwall:
All the new tests passed for me. On my system /tmp is not a tmpfs so I guess that explains it.

The testcases could create the chroot dir in the `tests` directory which is less likely to be a tmpfs and skip the test if it is a tmpfs.

stefanberger:
I opened PR #742 now to fix it.

PR #765, opened 2022-10-27
swtpm_setup: Configure swtpm to log to stdout/err if needed

rosslagerwall:
If swtpm_setup is configured with a log file, it launches swtpm configured with the same log file. If not, swtpm_setup logs will go to stdout/stderr and it should configure swtpm to do the same.

Signed-off-by: Ross Lagerwall <ross.lagerwall@citrix.com>

### DIFF
--- a/src/swtpm_setup/swtpm.c
+++ b/src/swtpm_setup/swtpm.c
@@ -66,6 +66,7 @@ static int swtpm_start(struct swtpm *self)
     struct stat statbuf;
     gboolean success;
     GError *error = NULL;
+    GSpawnFlags flags;
     unsigned ctr;
     int pidfile_fd;
     int ret = 1;
@@ -128,8 +129,16 @@ static int swtpm_start(struct swtpm *self)
     }
 #endif
 
-    success = g_spawn_async(NULL, argv, NULL,
-                            G_SPAWN_LEAVE_DESCRIPTORS_OPEN | G_SPAWN_STDOUT_TO_DEV_NULL | G_SPAWN_STDERR_TO_DEV_NULL,
+    flags = G_SPAWN_LEAVE_DESCRIPTORS_OPEN;
+    if (gl_LOGFILE) {
+        flags |= G_SPAWN_STDOUT_TO_DEV_NULL | G_SPAWN_STDERR_TO_DEV_NULL;
+    } else {
+#if GLIB_CHECK_VERSION(2, 74, 0)
+        flags |= G_SPAWN_CHILD_INHERITS_STDOUT | G_SPAWN_CHILD_INHERITS_STDERR;
+#endif
+    }
+
+    success = g_spawn_async(NULL, argv, NULL, flags,
                             NULL, NULL, &self->pid, &error);
     if (!success) {
         logerr(self->logfile, "Could not start swtpm: %s\n", error->message);


src/swtpm_setup/swtpm.c
```diff
@@ -128,8 +129,19 @@ static int swtpm_start(struct swtpm *self)
...
 #endif
 
-    success = g_spawn_async(NULL, argv, NULL,
-                            G_SPAWN_LEAVE_DESCRIPTORS_OPEN | G_SPAWN_STDOUT_TO_DEV_NULL | G_SPAWN_STDERR_TO_DEV_NULL,
+    flags = G_SPAWN_LEAVE_DESCRIPTORS_OPEN;
+    if (gl_LOGFILE) {
+        flags |= G_SPAWN_STDOUT_TO_DEV_NULL | G_SPAWN_STDERR_TO_DEV_NULL;
+    } else {
+#ifdef G_SPAWN_CHILD_INHERITS_STDOUT
+        flags |= G_SPAWN_CHILD_INHERITS_STDOUT;
```
stefanberger, line 137: The flags are typedefs and not #defines...  it won't set the flags then at all unless I am missing something. I suppose you did this to accommodate possibly older versions of glib-2.0 that don't have these flags?
```
typedef enum
{
  G_SPAWN_DEFAULT                = 0,
  G_SPAWN_LEAVE_DESCRIPTORS_OPEN = 1 << 0,
  G_SPAWN_DO_NOT_REAP_CHILD      = 1 << 1,
  /* look for argv[0] in the path i.e. use execvp() */
  G_SPAWN_SEARCH_PATH            = 1 << 2,
  /* Dump output to /dev/null */
  G_SPAWN_STDOUT_TO_DEV_NULL     = 1 << 3,
  G_SPAWN_STDERR_TO_DEV_NULL     = 1 << 4,
  G_SPAWN_CHILD_INHERITS_STDIN   = 1 << 5,
  G_SPAWN_FILE_AND_ARGV_ZERO     = 1 << 6,
  G_SPAWN_SEARCH_PATH_FROM_ENVP  = 1 << 7,
  G_SPAWN_CLOEXEC_PIPES          = 1 << 8
} GSpawnFlags;
```

stefanberger: It looks like `G_SPAWN_CHILD_INHERITS_STDOUT/STDERR` were introduced in 2.74: https://github.com/GNOME/glib/blob/bfe349983a7f32d969485cd29d8ce17e06121284/glib/gspawn.h#L191-L207

You may need to use the GLIB_CHECK_VERSION() macro in place of the two #ifdef's. (Example may be here: https://thrift.apache.org/tutorial/c_glib.html)

rosslagerwall: Oh, I missed that it was a #define. Will fix.

rosslagerwall:
I don't have a system with glib 2.74 to test this on but I think it should work.

stefanberger:
> I don't have a system with glib 2.74 to test this on but I think it should work now.

Why are you submitting patches that you haven't even tried out yourself?

stefanberger:
I tried it on Fedora Rawhide now.

stefanberger:
Thanks.

rosslagerwall:
> Why are you submitting patches that you haven't even tried out yourself?

I have built the patch and tested that it does the right thing, just not against glib 2.74 as I didn't have a distro with that version available. The `G_SPAWN_CHILD_INHERITS_{STDOUT,STDERR}` flag is implied by the default (no flag) behaviour anyway so the behaviour should be the same regardless.

Anyway, thanks for testing it on Fedora Rawhide.

PR #768, opened 2022-11-11
swtpm.spec: Reflect minimal gnutls version

zippy2:
The configure script requires at least gnutls-3.4.0 but this is not reflected in the spec file.

Signed-off-by: Michal Privoznik <mprivozn@redhat.com>


stefanberger:
Thanks.

PR #772, opened 2022-12-02
swtpm: Restore logging to stderr on log open failure

zippy2:
When opening a log file fails (--log file=/some/path) for whatever reason the @logfd variable is left set to -1 (the retval of open()). And due to how _logprintf() is written, the subsequent attempt to log this failure (logprintf()) returns too early not printing the useful error.

Therefore, restore the original (CONSOLE_LOGGING) value.

Signed-off-by: Michal Privoznik <mprivozn@redhat.com>


stefanberger:
Thank you!

PR #776, opened 2022-12-19
Couple of trivial fixes

zippy2:
These stem mostly from what fedora's spec file is doing. Except for the swtpm_cuse man page. Handling the installation of the file properly benefits other distros as well.

### DIFF
--- a/man/man8/Makefile.am
+++ b/man/man8/Makefile.am
@@ -24,8 +24,12 @@ man8_generated_MANS = \
 	swtpm-create-tpmca.8
 
 man8_static_MANS = \
-	swtpm_cuse.8 \
 	swtpm-localca.8
+
+if WITH_CUSE
+man8_static_MANS += \
+	swtpm_cuse.8
+endif
 
 man8_MANS = \
 	$(man8_generated_MANS) \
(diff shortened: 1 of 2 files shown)


stefanberger:
I saw this particular error only when building swtpm package on Fedora servers. There I added a patch for it temporarily since also my Fedora rawhide VM with the same gcc did not show it. My hope was it would go away since argv is initialized by a return value from a function right after all the variable declarations.

zippy2:
> I saw this particular error only when building swtpm package on Fedora servers. There I added a patch for it temporarily since also my Fedora rawhide VM with the same gcc did not show it. My hope was it would go away since argv is initialized by a return value from a function right after all the variable declarations.

Yeah, I was a bit hesitant to add this commit into the pull request. But then I figured, it's a step towards the safe side - if there's ever a conditional return introduced in between the variable declaration and its initialization we don't have to remember to initialize the variable. IOW, it's a good practice to have g_auto* variables initialized even though they are set right after the fact. But I can remove it, if you want me to.

stefanberger:
I just again updated my Fedora Rawhide VM and I am still not seeing the issue of the (mistaken) uninitialized variable. I take it now anyway.

PR #793, opened 2023-03-03
swtpm: Only set ancillary data on struct msghdr if file descriptor is domain AF_UNIX

stefanberger:


src/swtpm/utils.c
```diff
@@ -346,3 +348,21 @@ ssize_t read_eintr(int fd, void *buffer, size_t buflen)
...
+ * @fd: file descriptor of a socket
+ *
+ * Returns -1 in case an error occurred, the socket type otherwise
+ */
+int get_socket_domain(int fd)
+{
+    int optval;
+    socklen_t optlen = sizeof(optval);
+
+    if (getsockopt(fd, SOL_SOCKET, SO_DOMAIN, &optval, &optlen) < 0)
```
stefanberger, line 364: SO_DOMAIN is not a standard option supported on all platforms where swtpm runs on.

PR #794, opened 2023-03-09
Use PKG_CHECK_MODULES to check libtpms version

ben-dav-lytle:
`PKG_CHECK_MODULES` allows to check the library version (https://autotools.info/pkgconfig/pkg_check_modules.html - section 3.3), so there should be no need for the `AC_CHECK_LIB` call below it.

I was messing with macros in `TPMBuildSwitches.h` and would run into configure failing on https://github.com/stefanberger/swtpm/blob/459f4e7dea7320b2691dc6c69eb19a51787786fc/configure.ac#L185

libtpms was at `0.9.5` and `TPMLIB_ChooseTPMVersion` was present in `libtpms.so.*`. So that message was a little misleading.
I also found that with removing the `AC_CHECK_LIB` for `TPMLIB_ChooseTPMVersion`, the make process would tell me explicitly where an issue was. I had an undefined reference to either a variable or a function.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -176,16 +176,12 @@ AC_SUBST([LIBTASN1_LIBS])
 
 PKG_CHECK_MODULES(
 	[LIBTPMS],
-	[libtpms],
+	[libtpms >= 0.6],
 	,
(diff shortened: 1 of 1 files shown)


stefanberger:
Thanks. LGTM.

PR #811, opened 2023-08-02
Make stdout unbuffered in swtpm_{setup,localca}

rosslagerwall:
swtpm_setup and swtpm_localca may be run by a toolstack with log messages going to a pipe via stdout. Set stdout to be unbuffered to avoid log messages being lost if the process terminates unexpectedly.

### DIFF
--- a/src/swtpm_setup/swtpm_setup.c
+++ b/src/swtpm_setup/swtpm_setup.c
@@ -1230,6 +1230,8 @@ int main(int argc, char *argv[])
     int ret = 1;
     g_autoptr(GError) error = NULL;
 
+    setvbuf(stdout, 0, _IONBF, 0);
+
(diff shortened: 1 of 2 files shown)


src/swtpm_setup/swtpm_setup.c
```diff
@@ -1230,6 +1230,8 @@ int main(int argc, char *argv[])
     int ret = 1;
     g_autoptr(GError) error = NULL;
 
+    setvbuf(stdout, 0, _IONBF, 0);
```
stefanberger, line 1233: stderr is fine as it is?

rosslagerwall: Yes, `stderr` defaults to unbuffered.

stefanberger:
LGTM.

Issue #843, opened 2024-03-22
Disable test dependencies for normal build


jarkkojs:
Allow to build the image without socat and expect as they are required only for tests. This will make the project compatible with embedded build systems such as Yocto and BuildRoot.

jarkkojs:
Have not used Yocto for this, only BuildRoot but usually they have somewhat matching issues, so this should help in both environments.

jarkkojs:
OK, th

> > Have not used Yocto for this, only BuildRoot but usually they have somewhat matching issues, so this should help in both environments.
> 
> Yocto/meta-security already builds it: https://layers.openembedded.org/layerindex/recipe/319192/ I may add a `--without-test` option to configure.

A configure option would be good enough for BuildRoot. I.e. it does not require too much drift. So +1 for that.

stefanberger:
PR #844

jarkkojs:
> PR #844

Can you also point this out in the associated email thread in BuildRoot mailing list. I'll try this now and comment to the pull request, thanks!

PR #844, opened 2024-03-22
build-sys: Add support for --disable-tests to disable tests

stefanberger:
Add support for --disable-tests to disable the tests and the checking for tools that only need to be installed for the tests. Keep the tests enabled as default as it was before.

Resolves: https://github.com/stefanberger/swtpm/issues/843


jarkkojs:
It does work!

I created a test branch, which uses master.next and successfully run all kselftests for TPM2 using tpm_tis: https://gitlab.com/jarkkojs/linux-tpmdd-test/-/commits/swtpm-fix

PR #858, opened 2024-06-11
CHANGES: Add documentation for changes in 0.9.0

stefanberger:


CHANGES
```diff
@@ -1,8 +1,70 @@
 CHANGES - changes for swtpm
 
 version 0.9.0:
+  Note: The SElinux policy for swtpm was completely redone. For systems
+        with an SELinux policy the same policy as used in Fedora >= 40 is
```
elmarco, line 5: suggestion: which version of SELinux is now required

stefanberger: `>= 40.17`

PR #865, opened 2024-06-27
tests: Update to ibmtpm20tss v2.3.1 with rev183 tests

stefanberger:
Upgrade ibmtpm20tss test suite to v2.3.1 with rev183 tests -- wait for libtpms to support rev183.

If libtpms v0.10 with rev183 does not enable the new policy command (to keep full backwards compatibility of libtpms v0.10 with v0.9 if no profile support is added) then this patch may not be needed.


stefanberger:
@elmarco FYI: Since swtpm now requires libtpms >= 0.10 we can remove some patches from test using the ibmtss2 test suite...

elmarco:
yeah, I realized some patches were not needed since I have just applied a few of them in my installable-tests tree

PR #868, opened 2024-06-28
Update apparmor profile with additional permissions

lvoytek:
Add the sys_admin capability to the profile to allow access to kernel drivers like vtpm_proxy and add write permissions for non-owner files in /var/lib/libvirt/swtpm/ to fix lock file denials


stefanberger:
Thanks for the PR.

PR #870, opened 2024-07-09
Install tests

elmarco:
Allow tests/ to be installed and run from the system.

The installation path is loosely based on https://wiki.gnome.org/Initiatives/GnomeGoals/InstalledTests guidelines. We don't provide the .test files though, at this point. Instead, I wrote a simply installed-runner.sh script which mimics autotools basic runner. (ideally, the tests could use TAP instead etc)

Missing:
- [ ] running with SWTPM_TEST_EXPENSIVE=1
- [ ] SWTPM_TEST_STORE_VOLATILE=1 
- [ ] SWTPM_TEST_IBMTSS2=1
- [ ] and sudo

I think the current patch set should be ok for review/merge though.

### DIFF
--- a/tests/Makefile.am
+++ b/tests/Makefile.am
@@ -224,13 +224,15 @@ EXTRA_DIST = \
 	_test_tpm2_volatilestate \
 	_test_tpm2_wrongorder \
 	_test_volatilestate \
-	_test_wrongorder
+	_test_wrongorder \
+	installed-runner.sh
 
 testdir = $(libexecdir)/installed-tests/$(PACKAGE)
 
 install-data-hook:
 	mkdir -p $(DESTDIR)$(testdir)
 	cd $(srcdir) && for file in $(EXTRA_DIST); do install -D "$$file" "$(DESTDIR)$(testdir)/$$file" ; done
+	echo "$(TESTS)" > $(DESTDIR)$(testdir)/tests
 
 uninstall-hook:
 	cd $(DESTDIR)$(testdir) && rm -f $(EXTRA_DIST) tests
@@ -254,6 +256,6 @@ syntax-check:
 # SC2009: Consider using pgrep instead of grepping ps output.
 # SC2317: (info): Command appears to be unreachable. Check usage (or ignore if invoked indirectly).
 	shellcheck -e SC2009,SC2317 \
-		$(TESTS) $(TEST_UTILS) $(filter _test_%,$(EXTRA_DIST)) softhsm_setup
+		$(TESTS) $(TEST_UTILS) $(filter _test_%,$(EXTRA_DIST)) softhsm_setup installed-runner.sh
 
 check: check-am check-display
--- a/tests/Makefile.am
+++ b/tests/Makefile.am
@@ -226,6 +226,15 @@ EXTRA_DIST = \
 	_test_volatilestate \
 	_test_wrongorder
 
+testdir = $(libexecdir)/installed-tests/$(PACKAGE)
+
+install-data-hook:
+	mkdir -p $(DESTDIR)$(testdir)
+	cd $(srcdir) && for file in $(EXTRA_DIST); do install -D "$$file" "$(DESTDIR)$(testdir)/$$file" ; done
+
+uninstall-hook:
+	cd $(DESTDIR)$(testdir) && rm -f $(EXTRA_DIST) tests
+
 check-display:
 	@if test -n "$$SWTPM_EXE"; then \
 		echo "*** Using SWTPM_EXE=$$SWTPM_EXE"; \
--- a/tests/Makefile.am
+++ b/tests/Makefile.am
@@ -233,6 +233,7 @@ install-data-hook:
 	mkdir -p $(DESTDIR)$(testdir)
 	cd $(srcdir) && for file in $(EXTRA_DIST); do install -D "$$file" "$(DESTDIR)$(testdir)/$$file" ; done
 	echo "$(TESTS)" > $(DESTDIR)$(testdir)/tests
+	sed -i '/SWTPM_TEST_UNINSTALLED=1/d' $(DESTDIR)$(testdir)/common
 
 uninstall-hook:
 	cd $(DESTDIR)$(testdir) && rm -f $(EXTRA_DIST) tests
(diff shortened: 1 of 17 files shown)


tests/_test_tpm2_file_permissions
```diff
@@ -15,7 +15,6 @@ PIDFILE=${TPM_PATH}/swtpm.pid
 LOGFILE=${TPM_PATH}/swtpm.log
 PWDFILE=${TPM_PATH}/pwdfile.txt
 SWTPM_INTERFACE=${SWTPM_INTERFACE:-cuse}
-SWTPM_SETUP_CONF=${TPM_PATH}/swtpm_setup.conf
```
stefanberger, line 18: This one here will be dynamically created.

elmarco: ack

tests/common
```diff
@@ -8,12 +8,14 @@ if [ "${UNINSTALLED}" = "1" ]; then
     SWTPM_BIOS=${SWTPM_BIOS:-${ROOT}/src/swtpm_bios/swtpm_bios}
     SWTPM_SETUP=${SWTPM_SETUP:-${ROOT}/src/swtpm_setup/swtpm_setup}
     SWTPM_CERT=${SWTPM_CERT:-${ROOT}/src/swtpm_cert/swtpm_cert}
+    SWTPM_SETUP_CONF="$ROOT/samples/swtpm_setup.conf"
```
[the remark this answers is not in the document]
elmarco: ack

tests/Makefile.am
```diff
@@ -225,13 +225,15 @@ EXTRA_DIST = \
...
-	_test_wrongorder
+	_test_wrongorder \
+	installed-runner.sh
 
 testdir = $(libexecdir)/installed-tests/$(PACKAGE)
 
 install-data-hook:
 	mkdir -p $(DESTDIR)$(testdir)
 	cd $(srcdir) && cp --parents $(EXTRA_DIST) $(DESTDIR)$(testdir)
+	echo "$(TESTS)" > $(DESTDIR)$(testdir)/tests
```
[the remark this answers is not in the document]
elmarco: fixed

tests/Makefile.am
```diff
@@ -227,6 +227,12 @@ EXTRA_DIST = \
 	_test_volatilestate \
 	_test_wrongorder
 
+testdir = $(libexecdir)/installed-tests/$(PACKAGE)
+
+install-data-hook:
+	mkdir -p $(DESTDIR)$(testdir)
+	cd $(srcdir) && cp --parents $(EXTRA_DIST) $(DESTDIR)$(testdir)
```
stefanberger, line 235: `make distcheck` doesn't complete as normal user anymore...

elmarco: Fun, `cp --parents` creates the directory for the first file and then sets the directory read-only, like the source. Other file copy fails :) fixed

tests/Makefile.am
```diff
@@ -9,7 +9,8 @@ check_PROGRAMS =
 TESTS_ENVIRONMENT = \
   abs_top_testdir=`cd '$(top_srcdir)'/tests; pwd` \
   abs_top_builddir=`cd '$(top_builddir)'; pwd` \
-  abs_top_srcdir=`cd '$(top_srcdir)'; pwd`
+  abs_top_srcdir=`cd '$(top_srcdir)'; pwd` \
+  UNINSTALLED=1
```
stefanberger, line 13: I would prefer to not having to set UNINSTALLED=1 for every test I run from the command line. I think your test running should set INSTALLED=1.

stefanberger:
The problem with the pkcs11 test failing seems to be due to the PATH being set in the test case not pointing to the correct directory anymore. It looks like this:

+ PATH=/src/swtpm_cert:/usr/local/bin:/usr/bin:/usr/local/sbin:/usr/sbin

tests/Makefile.am
elmarco: ok

tests/common
```diff
@@ -1,12 +1,21 @@
...
+    SWTPM_BIOS=${SWTPM_BIOS:-${ROOT}/src/swtpm_bios/swtpm_bios}
+    SWTPM_SETUP=${SWTPM_SETUP:-${ROOT}/src/swtpm_setup/swtpm_setup}
+    SWTPM_CERT=${SWTPM_CERT:-${ROOT}/src/swtpm_cert/swtpm_cert}
+else
+    SWTPM_EXE=${SWTPM_EXE:-$(type -P swtpm)}
+    SWTPM_IOCTL=${SWTPM_IOCTL:-$(type -P swtpm_ioctl)}
+    SWTPM_BIOS=${SWTPM_BIOS:-$(type -P swtpm_bios)}
+    SWTPM_SETUP=${SWTPM_SETUP:-$(type -P swtpm_setup)}
+    SWTPM_CERT=${SWTPM_CERT:-$(type -P swtpm_cert)}
+fi
```
stefanberger, line 21: All of these could go into the test runner and the changes to this file could be removed.

stefanberger (changes requested):

tests/common
elmarco: kinda, it's unfortunate not to be able to run the test individually easily.. Now you have to run them with INSTALLED=1, which is kinda ok until we have a better solution...alternatively, we generate a different file for installation. wdyt?

stefanberger: There's test_config.in that could be sourced by all tests and contain INSTALLED=0|1 (maybe edited with sed when installing?)

elmarco: See last patch for a nicer alternative.

stefanberger: LGTM. Call variable `SWTPM_TEST_UNINSTALLED`?

elmarco: done

stefanberger:
If you remove SWTPM_EXE from being set in test_tpm2_vtpm_proxy, test_vtpm_proxy, test_tpm2_save_load_state_2, test_tpm2_save_load_state_3 (and SWTPM_IOCTL from latter two) as well as a few others, then these work in the installed case with sudo.
`SWTPM_TEST_EXPENSIVE=1 ./installed-runner.sh` seems to already do what it is supposed to.

stefanberger:
Can you squash all the 'teach' patches?

stefanberger:
2nd PR for these here?
```
SKIP: test_vtpm_proxy
SKIP: test_tpm2_vtpm_proxy
SKIP: test_ctrlchannel2
SKIP: test_tpm2_chroot_socket
SKIP: test_tpm2_chroot_chardev
SKIP: test_tpm2_chroot_cuse
SKIP: test_tpm2_ctrlchannel2
SKIP: test_tpm2_partial_reads
SKIP: test_tpm2_save_load_state_2
SKIP: test_tpm2_save_load_state_2_linear
SKIP: test_tpm2_save_load_state_2_block
SKIP: test_tpm2_save_load_state_3
```

elmarco:
Yes I plan to do the rest as follow-ups

PR #873, opened 2024-07-16
Enable extra installed tests

elmarco:


tests/test_tpm2_ibmtss2
```diff
@@ -44,77 +44,80 @@ SWTPM_SERVER_NO_DISCONNECT="1" run_swtpm "${SWTPM_INTERFACE}" \
 	--tpmstate "dir=${WORKDIR}" \
 	--flags not-need-init
 
+revision=$(run_swtpm_ioctl "${SWTPM_INTERFACE}" --info 1 |
```
stefanberger, line 47: So for this patch we have to wait for the merge request to make it? Maybe you should separate-out this patch so we can merge the other two?

elmarco: Maybe we don't have to, it will fallback to install mode if it doesn't find the tests installed. I didn't get feedback from my PR yet https://sourceforge.net/p/ibmtpm20tss/tss/merge-requests/2/, maybe you could comment there? Also I wonder if we can have those patches in RHEL/Fedora in the meantime.

stefanberger: With this test case relying on an installed version of the TSS 2 stack and its test suite libtpms + swtpm get entangled with it. So if the test suite changes in some form and tests something that's not supported by libtpms at that point we get an error. And then we may see these problems across distros as well? I don't like this.

elmarco: Is it better than not running those tests though?

stefanberger: > Is it better than not running those tests though?

I don't want Bugzillas across distros because of this.

elmarco: Well, those would be distos issues then. As long as you can say upstream this set of packages work together.

stefanberger: libtpms may be lagging behind tss2 by weeks or months and we cannot force distros to downgrade. So I don't really want this potential source of BZs.

elmarco: If you think it's necessary, we could have an extra environemnt variable to run against the installed tests

stefanberger: I think we have to gate it somehow. If tests don't pass on a machine just because the TSS2 tests are too new then people don't know what's wrong. So yes, I think we need some environment variable and a warning in the tests.

elmarco: I added INSTALLED_TESTS_IBMTSS env, is that ok with it?

stefanberger: I think such a variable should have prefix `SWTPM_TEST_` .

elmarco: done

stefanberger:
Can you please rebase.

All of these work fine (run as non-root):
```
$ ./installed-runner.sh
$ SWTPM_TEST_EXPENSIVE=1 ./installed-runner.sh
$ SWTPM_TEST_EXPENSIVE=1 SWTPM_TEST_IBMTSS=1 ./installed-runner.sh
```

I saw test_tpm12 failing due to the following but then also passing at least once -- probably the git clone didn't work:
```
./test_tpm12: line 80: ./patches/libtpm.patch: No such file or directory
```

The following tests are still being skipped:
```
SKIP: test_tpm2_save_load_state_2
SKIP: test_tpm2_save_load_state_2_linear
SKIP: test_tpm2_save_load_state_2_block
SKIP: test_tpm2_save_load_state_3
```

elmarco:
> The following tests are still being skipped:

You still need SWTPM_TEST_IBMTSS2=1.

To run against an installed IBM TSS test suite, you may set the
`SWTPM_TEST_IBMTSS` to the location of the test suite, such as
`/usr/libexec/installed-tests/ibmtss`.

stefanberger:
#892 should go in before this one.

elmarco:
> #892 should go in before this one.

there isn't much left after the rebase, thanks

PR #874, opened 2024-07-17
Enable setting the default RSA keysize using swtpm_setup.conf

stefanberger:
This PR adds the ability to provide a default RSA keysize using swtpm_setup.conf. It is used unless the user provided it with command line option to swtpm_setup.


stefanberger:
@elmarco

PR #875, opened 2024-07-18
Adding dockerfile and ci/cd

### LINKED ISSUE #869: Any objection if I add Dockerfile here ?
wiki page https://github.com/stefanberger/swtpm/wiki/Dockerfiles is not enough...

something like this is super easy and helpful:

```
$ cat Dockerfile
FROM debian:12.6-slim

RUN apt-get update && \
    apt-get install --no-install-recommends -y swtpm=0.7.1-1.3 && \
    apt-get clean && \
    rm -rf /var/lib/apt/lists/*

ENTRYPOINT ["/usr/bin/swtpm"]
```
lint passed:
```
$ docker run --rm -i ghcr.io/hadolint/hadolint < Dockerfile
$
```

build:
```
$ docker build -t swtpm:0.7.1 - < Dockerfile
Sending build context to Docker daemon  2.048kB
Step 1/3 : FROM debian:12.6-slim
 ---> 3eeb141ea1a1
Step 2/3 : RUN apt-get update &&     apt-get install --no-install-recommends -y swtpm=0.7.1-1.3 &&     apt-get clean &&     rm -rf /var/lib/apt/lists/*
 ---> Using cache
 ---> a9e31f6fdb72
Step 3/3 : ENTRYPOINT ["/usr/bin/swtpm"]
 ---> Using cache
 ---> 6fb7ab4b45f6
Successfully built 6fb7ab4b45f6
Successfully tagged swtpm:0.7.1
```
run:
```
$ docker run --rm -it swtpm:0.7.1 --version
TPM emulator version 0.7.1, Copyright (c) 2014-2021 IBM Corp.
```

glimchb:
Fixes #869

- Adding alpine based dockerfile
- ci: adding linter and builder github action
- ci: add testing for dockerfile
- ci: add new release action to run when published

### DIFF
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -1,0 +1,75 @@
+name: Docker
+
+on:
+  workflow_dispatch:
+  workflow_call:
+  push:
+    branches: [ "main" ]
+  pull_request:
+    branches: [ "main" ]
+
+jobs:
+
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: hadolint/hadolint-action@v3.1.0
+        with:
+          recursive: true
+          ignore: DL3018
+
+  build:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
+      # This is used to complete the identity challenge
+      # with sigstore/fulcio when running outside of PRs.
+      id-token: write
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      # Workaround: https://github.com/docker/build-push-action/issues/461
+      - name: Setup Docker buildx
+        uses: docker/setup-buildx-action@v3.0.0
+
+    #   - name: Log in to Docker Hub
+    #     if: github.event_name != 'pull_request'
+    #     uses: docker/login-action@v3.0.0
+    #     with:
+    #       username: ${{ secrets.DOCKERHUB_USERNAME }}
+    #       password: ${{ secrets.DOCKERHUB_TOKEN }}
+
+      - name: Log in to the Container registry
+        if: github.event_name != 'pull_request'
+        uses: docker/login-action@v3.0.0
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      # Extract metadata (tags, labels) for Docker
+      # https://github.com/docker/metadata-action
+      - name: Extract Docker metadata
+        id: meta
+        uses: docker/metadata-action@v5.5.0
+        with:
+          images: |
+            ghcr.io/${{ github.repository }}
+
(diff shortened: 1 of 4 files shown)


.github/workflows/release.yaml
```diff
@@ -0,0 +1,12 @@
+name: Release
+
+on:
+    workflow_dispatch:
+    release:
+        types: [published]
+
```
stefanberger, line 7: So this will only push once v0.10 is release?

glimchb: `.github/workflows/docker.yaml` will push after afery PR merged
`.github/workflows/release.yaml` will push on every release with a tag

stefanberger (approved):

.github/workflows/docker.yaml
```diff
@@ -0,0 +1,104 @@
...
+      # Workaround: https://github.com/docker/build-push-action/issues/461
+      - name: Setup Docker buildx
+        uses: docker/setup-buildx-action@v3.0.0
+
+    #   - name: Log in to Docker Hub
+    #     if: github.event_name != 'pull_request'
+    #     uses: docker/login-action@v3.0.0
+    #     with:
+    #       username: ${{ secrets.DOCKERHUB_USERNAME }}
+    #       password: ${{ secrets.DOCKERHUB_TOKEN }}
```
stefanberger, line 44: I now set these two as repository secrets. So theoretically this should work.

stefanberger: I am not sure how to enable this file here and see it running already as part of the PR...

glimchb: I will enable this in next patch, no problem, once you added secrets, should be easy for me to add this

glimchb:
@stefanberger can you merge this , please ? Travis CI passed

stefanberger:
Thanks!

.github/workflows/docker.yaml
glimchb: opened #876

stefanberger:
I had to fix the Dockerfile for github ci/cd since it started to fail due to an update of libtpms adding a new API call that then wasn't available to swtpm because the libtpms build didn't happen and the old version of libtpms was used: PR #898  
There were lots of broken github builds for the build action lately. I wouldn't mind it skipping the libtpms build for as long as it hasn't been updated but when an update happens it should build again. I suppose there's no way of doing this easily...

glimchb:
@stefanberger one way to do it easily is add `libtpms` as a git submodule of `swtpm`
```
cd swtpm
git submodule add ../libtpms
```
in this way every time to checkout `swtpm` it will bring correct `libtpms` with it

so docker build and all other builds, will just do normal checkout of `swtpm` and don;t need to worry about `libtpm`

every time you need to bump the `libtpm`, you just checkout and commit as usual...

WDYT ?


another option - I exposed `LIBTPMS_BRANCH` argument, so we can just pass specific commit/tag instead of master...

another option - we use `--no-cache` in docker build... so it will always use latest master...

PR #878, opened 2024-07-22
DO-NOT-MERGE: Disable OpenSSL's FIPS mode if enabled algorithms require it

stefanberger:
When FIPS mode is enabled in OpenSSL then certain algorithms are disabled. To still allow a swtpm instance to run under these circumstances, check for whether a certain set of algorithms are part of the  swtpm instance's profile and if so disable FIPS mode in the OpenSSL instance.


stefanberger:
@elmarco @berrange These are for now the unchanged remainders of the original profiles PR. I believe swtpm still needs to check whether FIPs mode is enabled in OpenSSL but maybe display an error message if it is and algorithms are not functioning and exit, OR allow to disable FIPS mode in the OpenSSL instance if the user passed a command line option.

stefanberger:
It seems there are different ways we could disable the enforcement of disabled algorithms on a host:
- OPENSSL_no_config() : do not read any configuration file; very last resort
- EVP_default_properties_enable_fips(NULL, 0): this seems to enable a few algorithms disabled by FIPS mode; `/etc/crypto-policies/back-ends/opensslcnf.config` not changed between fips and non-fips mode

My current thinking is to 

1. check for FIPS mode enabled and 
2. if it is enabled test candidate algorithms to be disabled by FIPS mode 
3. then disable FIPS mode if necessary and 
4. then test the same candidate algorithms again and possibly more 
5. and exit with error if it fails


This is a test program:

```
#include <openssl/evp.h>
#include <openssl/conf.h>

int main(void)
{
//    OPENSSL_no_config();
//    EVP_default_properties_enable_fips(NULL, 1);

    printf("AES-128-CFB: %d\n", EVP_CIPHER_fetch(NULL, "AES-128-CFB", NULL));
    printf("AES-256-CFB: %d\n", EVP_CIPHER_fetch(NULL, "AES-256-CFB", NULL));
    printf("CAMELLIA-128-CFB: %d\n", EVP_CIPHER_fetch(NULL, "CAMELLIA-128-CFB", NULL));
    printf("CAMELLIA-256-CFB: %d\n", EVP_CIPHER_fetch(NULL, "CAMELLIA-256-CFB", NULL));
    printf("DES-EDE3-CFB: %d\n", EVP_CIPHER_fetch(NULL, "DES-EDE3-CFB", NULL));

    EVP_default_properties_enable_fips(NULL, 0);

    printf("AES-128-CFB: %d\n", EVP_CIPHER_fetch(NULL, "AES-128-CFB", NULL));
    printf("CAMELLIA-128-CFB: %d\n", EVP_CIPHER_fetch(NULL, "CAMELLIA-128-CFB", NULL));
    printf("CAMELLIA-256-CFB: %d\n", EVP_CIPHER_fetch(NULL, "CAMELLIA-256-CFB", NULL));
    printf("DES-EDE3-CFB: %d\n", EVP_CIPHER_fetch(NULL, "DES-EDE3-CFB", NULL));
}
```

On a FIPS-enabled F40 host this prints:
```
AES-128-CFB: 1005757168
AES-256-CFB: 1005756496
CAMELLIA-128-CFB: 0
CAMELLIA-256-CFB: 0
DES-EDE3-CFB: 0
AES-128-CFB: 1005653632
CAMELLIA-128-CFB: 1005727792
CAMELLIA-256-CFB: 1005725840
DES-EDE3-CFB: 1005741520
```

PR #890, opened 2024-08-21
tests: Set OPENSSL_ENABLE_SHA1_SIGNATURES=1 for IBMTSS2 test

stefanberger:
The IBMTSS2 tests uite creates signatures over SHA1 that may now fail on RHEL 9.x and CentOS 9. To have these tests succeed set OPENSSL_ENABLE_SHA1_SIGNATURES=1 so the tests do not need to be modified and also check that the TPM 2 can handle SHA1 signatures as before. 'swtpm socket --tpm2' should set this environment variable automatically if needed.

### DIFF
--- a/tests/test_tpm2_ibmtss2
+++ b/tests/test_tpm2_ibmtss2
@@ -104,7 +104,7 @@ if ! ./startup; then
 	exit 1
 fi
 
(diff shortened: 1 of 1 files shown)


stefanberger:
@elmarco : FYI

PR #891, opened 2024-08-22
swtpm: Modify custom profile when FIPS is enabled on host

stefanberger:
When FIPS is enabled on the host and the 'custom' profile is chosen then remove all (currently) known algorithms disabled by FIPS so that FIPS does not need to be disabled in the OpenSSL instance. Also set or adjust minimum key sizes for EC and RSA keys.


stefanberger:
A few questions:
- Is this patch needed? Or should we rely on higher layers (libvirt, users) to provide appropriate profiles when FIPS mode is enabled? It's easier to deal with removal of algorithms from custom profile on swtpm layer that's for sure.
- Should more algorithms be removed from the custom profile, such as those mentioned in [FIPS-140-3 guidance document](https://trustedcomputinggroup.org/wp-content/uploads/TCG-FIPS-140-3-Guidance-for-TPM-2.0-Version-1.0-Revision-2_pub-1.pdf)? 
  - section 5.6: ecschnorr, ecdaa, ecc-bn, ecc-bn-p*, ecc-sm2-p*

PR #892, opened 2024-08-26
Allow more tests to run installed

stefanberger:
This PR cleans up the test cases so they all use the SWTPM_XYZ defines from file `common` so that all test cases can run 'installed'.


stefanberger:
@elmarco FYI

PR #898, opened 2024-09-03
Fix the building of the Docker container on github

stefanberger:


glimchb:
you can also use `no-cache` option in the workflow https://github.com/stefanberger/swtpm/blob/607eb54b3e52a8c6fd90676f12e62ec6f8320f61/.github/workflows/docker.yaml#L76

PR #932, opened 2024-10-14
Add support for file-backend locking

stefanberger:
This PR adds support for optional file-backend locking using a new lock option parameter that is to be used with the --tpmstate option.


stefanberger:
@elmarco @berrange This PR adds file-backend storage locking.

elmarco:
LGTM, although patches ordering and atomicity isn't optimal, np

stefanberger:
> LGTM, although patches ordering and atomicity isn't optimal, np

Reordered the patches. With cli option support coming now after the enablement of file backend locking this should take care of the issue.Will merge soon.

I haven't used this backend much. So needs some extra testing in libvirt environment with suspend/resume etc..

PR #936, opened 2024-10-24
Add support for recreate-svn-base-secret flag and call LIBTPM_RecreateSvnBaseSecret

stefanberger:


stefanberger:
This PR depends on https://github.com/stefanberger/libtpms/pull/470

PR #940, opened 2024-10-28
test: Exit IBM TSS2 test early if it does not support swtpm

stefanberger:
Check the help screen for necessary supported options since the IBM TSS2 test will have to be patched to support swtpm directly. If it does not support it, exit the tests early with an error message.


elmarco:
ok, thanks. Fwiw, Ken no longer replies to my MR https://sourceforge.net/p/ibmtpm20tss/tss/merge-requests/2/...

PR #941, opened 2024-10-31
configure.ac: enable 64-bit file API on 32-bit systems

trofi:
My local filesystem is btrfs with a long life. It's inodes ecxeed 32-bit
space and that causes test failures in `swtpm` on `i686-linux`
containers:

    FAIL: test_parameters
    FAIL: test_swtpm_setup_file_backend
    FAIL: test_swtpm_setup_overwrite
    FAIL: test_tpm2_swtpm_setup_create_cert
    FAIL: test_tpm2_swtpm_setup_overwrite
    FAIL: test_swtpm_setup_create_cert
    FAIL: test_tpm2_parameters

The example test failure log looks this way:

    FAIL: test_migration_key
    ========================

    Need to be root to run test with CUSE interface.
    Need to be root to run test with CUSE interface.
    ==== Starting swtpm with interfaces socket+socket ====
    Test 1: Ok
    ==== Starting swtpm with interfaces socket+socket ====
    Test 2: Ok
    ==== Starting swtpm with interfaces socket+socket ====
    swtpm: Missing migration key to decrypt volatilestate
    Test 3: Ok
    ==== Starting swtpm with interfaces socket+socket ====
    Could not stat file '/build/tests/data/migkey1/volatilestate.bin': Value too large for defined data type
    Error: Could not load encrypted volatile state into TPM.
    FAIL test_migration_key (exit status: 1)

The `stat()` fails because inode value exceeds 32-bit value:

    $ stat /build/tests/data/migkey1/volatilestate.bin
      File: /build/tests/data/migkey1/volatilestate.bin
      Size: 1290            Blocks: 8          IO Block: 4096   regular file
    Device: 0,30    Inode: 9639547569  Links: 1
    ...

The change fixes all the test failures. To fix
`test_tpm2_swtpm_setup_create_cert` I also had to include `config.h`
into `swtpm_backend_dir.c` to get 64-bit file open there as well.

Signed-off-by: Sergei Trofimovich <slyich@gmail.com>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -48,6 +48,10 @@ AC_CANONICAL_HOST
 AM_INIT_AUTOMAKE([foreign 1.6])
 AM_SILENT_RULES([yes])
 
+# Allow 64-bit file API on 32-bit systems. Without the change even small
+# files will fail to stat any files on filesystems with 64-bit inodes.
+AC_SYS_LARGEFILE
+
 DEBUG=""
 AC_MSG_CHECKING([for debug-enabled build])
 AC_ARG_ENABLE(debug, AS_HELP_STRING([--enable-debug],[create a debug build]),
--- a/src/swtpm_setup/swtpm_backend_dir.c
+++ b/src/swtpm_setup/swtpm_backend_dir.c
@@ -6,6 +6,8 @@
  * Refactored as module: Stefan Reiter, stefan@pimaker.at
  */
 
+#include "config.h"
+
 #include <errno.h>
 #include <string.h>
 #include <unistd.h>


configure.ac
```diff
@@ -48,6 +48,9 @@ AC_CANONICAL_HOST
 AM_INIT_AUTOMAKE([foreign 1.6])
 AM_SILENT_RULES([yes])
 
+# allow 64-bit file API on 32-bit systems.
```
stefanberger, line 51: commit message: s/ilfe/life

Otherwise sounds and looks good.

trofi: Updated. Thank you!

stefanberger: Can you also explain a bit what the errors are related to and where they occur? We don't have any large files with swtpm, so this is not quite so obvious. Can you provide this detail, please?

trofi: Updated as:

```diff
--- a/configure.ac
+++ b/configure.ac
@@ -51 +51,2 @@ AM_SILENT_RULES([yes])
-# allow 64-bit file API on 32-bit systems.
+# Allow 64-bit file API on 32-bit systems. Without the change even small
+# files will fail to open any files on filesystems with 64-bit inodes.
```

Or should I expand the commit message itself with breakage specifics?

stefanberger: No, I think this is good.
Though I am trying to understand what is causing the issue. Do all applications have to use the 64bit file API  on your type of system? What matters for swtpm are filenames, flags, and file descriptors. The 64bit inodes presumably are entirely handled in the kernel. So I fail to see the connection.

baloo: I think that's the specific bit it fixes:
https://github.com/autotools-mirror/autoconf/blob/f8c82d292699fbce6d60abb46259a3781578f7fc/lib/autoconf/specific.m4#L343

I didn't check but swtpm probably issue a `stat(2)` and gets an error because the inode overflows.

I would guess this is a broader issue on @trofi's system and that every app needs this fix.
I don't think the macro hurts much as autotool will probe to check if the system supports it.

baloo: although, what autotools does is adding `-D_FILE_OFFSET_BITS` (https://github.com/autotools-mirror/autoconf/blob/f8c82d292699fbce6d60abb46259a3781578f7fc/lib/autoconf/specific.m4#L322) but glibc has had a fix for that for a fairly long time:
https://github.com/bminor/glibc/blame/9c0903fb7388f645d23b26160ed3669a116189fe/bits/stat.h#L36-L40

stefanberger: > I didn't check but swtpm probably issue a stat(2) and gets an error because the inode overflows.

swtpm use `fstat` and `stat` and I would expect that several more test should fail if these wwere to run into issues. Particularly those tests that use existing state and start swtpm with it and swtpm will do a fstat on it to determine file size.

> Or should I expand the commit message itself with breakage specifics?

@trofi Actually if you could check the log of the failing tests and if there's something of interest in there, and there hopefully is, then copy and paste it into the commit message.

trofi: Updated PR description to contain example failure.

> I would guess this is a broader issue on @trofi's system and that every app needs this fix.

Yes, newly created files on my system have inode values higher than 2^32. It's a good way to expose failures like this one.

> but glibc has had a fix for that for a fairly long time

At a hazard of stating the obvious: the `glibc` fix is not automatic as it's a visible ABI change. Effectively every single 32-bit application has to be built with `-D_FILE_OFFSET_BITS=64` to be able to work on the filesystems that can expose 64-bit values to userspace. Unfortunately it changes linkage symbols from `stat()` to `stat64()` and breaks some programs that don't expect the symbol change and is thus not enabled by default in `glibc`. Fun fact: `meson` does enable 64-bit file API by default: https://github.com/mesonbuild/meson/blob/4ac3e7d356cfc1a6a290cd6961f01c11510c0618/mesonbuild/compilers/compilers.py#L1121

configure.ac
```diff
@@ -48,6 +48,10 @@ AC_CANONICAL_HOST
 AM_INIT_AUTOMAKE([foreign 1.6])
 AM_SILENT_RULES([yes])
 
+# Allow 64-bit file API on 32-bit systems. Without the change even small
+# files will fail to open any files on filesystems with 64-bit inodes.
```
stefanberger, line 52: fail to 'stat'?

trofi: Updated as:

```
--- a/configure.ac
+++ b/configure.ac
@@ -52 +52 @@ AM_SILENT_RULES([yes])
-# files will fail to open any files on filesystems with 64-bit inodes.
+# files will fail to stat any files on filesystems with 64-bit inodes.
```

stefanberger:
Thanks. I will merge it now.

PR #944, opened 2024-11-05
debian: Add rule to allow usage of /var/tmp directory (QEMU)

stefanberger:
QEMU's functional tests need access to /var/tmp/**. To avoid the following
type of AppArmor permission failures add a rule that allows access to
/var/tmp/**.

 type=AVC msg=audit(1730829888.863:260): apparmor="DENIED" \
   operation="mknod" class="file" profile="swtpm" \
   name="/var/tmp/qemu_3r9txw7z/swtpm-socket" pid=3925 comm="swtpm" \
   requested_mask="c" denied_mask="c" fsuid=1000 ouid=1000FSUID="stefanb" \
   OUID="stefanb"

[ To run the QEMU's functional tests use the following command:
    make check-functional ]

### DIFF
--- a/debian/usr.bin.swtpm
+++ b/debian/usr.bin.swtpm
@@ -4,6 +4,7 @@
 #include <tunables/global>
 
 profile swtpm /usr/bin/swtpm {
+  #include <abstractions/user-tmp>
   #include <abstractions/base>
   #include <abstractions/openssl>
 


stefanberger:
FYI, @lvoytek . Let me know if you have a comment...

lvoytek:
> FYI, @lvoytek . Let me know if you have a comment...

Looks good to me, tmp access should be fine and the user-tmp abstraction is a good way to add it. The commit title says /var/lib directory though

stefanberger:
> > FYI, @lvoytek . Let me know if you have a comment...
> 
> Looks good to me, tmp access should be fine and the user-tmp abstraction is a good way to add it. The commit title says /var/lib directory though

Thanks for the feedback. I fixed the title.

stefanberger:
@lvoytek I now file this bug for Ubuntu: https://bugs.launchpad.net/ubuntu/+source/swtpm/+bug/2086736

stefanberger:
@lvoytek Actually, the original bug report show this audit entry:

```
audit(1730822474.384:446): apparmor="DENIED" operation="mknod"
profile="swtpm"
name="/mnt/nvmedisk/linaro/qemu-from-laptop/qemu/build/tests/functional/arm/test_arm_aspeed.AST2x00Machine.test_arm_ast2600_evb_buildroot_tpm/qemu-machine-hhuvwytc/.lock"
pid=2820156 comm="swtpm" requested_mask="c" denied_mask="c" fsuid=1000
ouid=1000
```

So they are using `/mnt`, for which we would probably need a rule like `owner /mnt/** rwkl` or even `owner /** rwkl`.

lvoytek:
> @lvoytek Actually, the original bug report show this audit entry:
> 
> ```
> audit(1730822474.384:446): apparmor="DENIED" operation="mknod"
> profile="swtpm"
> name="/mnt/nvmedisk/linaro/qemu-from-laptop/qemu/build/tests/functional/arm/test_arm_aspeed.AST2x00Machine.test_arm_ast2600_evb_buildroot_tpm/qemu-machine-hhuvwytc/.lock"
> pid=2820156 comm="swtpm" requested_mask="c" denied_mask="c" fsuid=1000
> ouid=1000
> ```
> 
> So they are using `/mnt`, for which we would probably need a rule like `owner /mnt/** rwkl` or even `owner /** rwkl`.

That's fair, I guess it would be reasonable to allow swtpm full access to any file that it specifically owns. We would also be able to clean up the apparmor profile a lot by removing all the lines tagged with owner, and it would likely prevent a lot of future unnecessary apparmor denials. I'll check with my team to make sure that would be fine to add into Ubuntu.

stefanberger:
> We would also be able to clean up the apparmor profile a lot by removing all the lines tagged with owner, and it would likely prevent a lot of future unnecessary apparmor denials. I'll check with my team to make sure that would be fine to add into Ubuntu.

Yes, if this is indeed considered a plus. All file access restrictions that swtpm once had due to the profile will be gone...

stefanberger:
> > We would also be able to clean up the apparmor profile a lot by removing all the lines tagged with owner, and it would likely prevent a lot of future unnecessary apparmor denials. I'll check with my team to make sure that would be fine to add into Ubuntu.
> 
> Yes, if this is indeed considered a plus. All file access restrictions that swtpm once had due to the profile will be gone...

... though we may not have another choice. Next time root could mount to `/mymount` and we'd have the same problem again. :-(

lvoytek:
> > > We would also be able to clean up the apparmor profile a lot by removing all the lines tagged with owner, and it would likely prevent a lot of future unnecessary apparmor denials. I'll check with my team to make sure that would be fine to add into Ubuntu.
> > 
> > 
> > Yes, if this is indeed considered a plus. All file access restrictions that swtpm once had due to the profile will be gone...
> 
> ... though we may not have another choice. Next time root could mount to `/mymount` and we'd have the same problem again. :-(

Going over it with my team, we ultimately decided that /** with even owner files would still provide too much access, since a change of file/directory uid to swtpm would ultimately remove any intended restrictions. For mount directory use cases we would recommend adding a local override of the folder name if it's known. I still think having all tmp directories available by default would be reasonable, but I know it doesn't solve this specific issue :(

stefanberger:
> Going over it with my team, we ultimately decided that /** with even owner files would still provide too much access, since a change of file/directory uid to swtpm would ultimately remove any intended restrictions. For mount directory use cases we would recommend adding a local override of the folder name if it's known. I still think having all tmp directories available by default would be reasonable, but I know it doesn't solve this specific issue :(

I am going to try to adjust the test case and send a patch to the QEMU mailing list. The path in the test case is not ideal. However, we will still need this patch here, since it will be using `/var/tmp` then.

stefanberger:
Proposed QEMU patch: https://lore.kernel.org/qemu-devel/20241106180751.6859-1-stefanb@linux.vnet.ibm.com/T/#u

Still need this one here.

PR #953, opened 2024-11-14
Fix some issues when building an rpm from swptm.spec

stefanberger:


stefanberger:
@elmarco I added a swtpm-tests package to the sample spec file and was also going to do that on Fedora rather than putting the tests into an existing package. Should be ok?

elmarco:
> @elmarco I added a swtpm-tests package to the sample spec file and was also going to do that on Fedora rather than putting the tests into an existing package. Should be ok?

yes, that's what I would have done. thanks

PR #957, opened 2024-12-10
swtpm_setup.c: add "socket" argument to custom swtpm call

mikkorapeli-linaro:
When a custom swtpm binary is provided with --tpm path/to/swtpm, then swtpm_setup is currently not adding "swtpm socket" argument when calling swtpm --print-capabilities which results in swtpm_setup always failing. --tpm documentation does not mention that "socket" argument should be provided so add it automatically. As workaround, --tpm "path/to/swtpm socket" works but it should be the default just like when swtpm is found from PATH.

Fixes use of swtpm in yocto cross compile environment where paths to swtpm and swtpm_setup binaries provided by swtpm-native recipe change frequently due to
recipe specific sysroot environment.

### DIFF
--- a/src/swtpm_setup/swtpm_setup.c
+++ b/src/swtpm_setup/swtpm_setup.c
@@ -1472,6 +1472,11 @@ int main(int argc, char *argv[])
         case 'T': /* --tpm */
             g_free(swtpm_prg);
             swtpm_prg = g_strdup(optarg);
+            if (swtpm_prg) {
+                tmp = g_strconcat(swtpm_prg, " socket", NULL);
+                g_free(swtpm_prg);
+                swtpm_prg = tmp;
+            }
             break;
         case '_': /* --swtpm_ioctl */
             fprintf(stdout, "Warning: --swtpm_ioctl is deprecated and has no effect.");


src/swtpm_setup/swtpm_setup.c
```diff
@@ -1472,6 +1472,11 @@ int main(int argc, char *argv[])
         case 'T': /* --tpm */
             g_free(swtpm_prg);
             swtpm_prg = g_strdup(optarg);
+            if (swtpm_prg) {
+                tmp = g_strconcat(swtpm_prg, " socket", NULL);
+                g_free(swtpm_prg);
```
stefanberger, line 1477: This should not be necessary. Test cases already pass it like this using quotation marks around the parameters "swtpm socket" :
```
                             "./${libtpmsdir}/swtpm_setup" \
                                --tpm2 \
                                --tpmstate "${statedir}" \
                                --config "${workdir}/swtpm_setup.conf" \
                                --log "${statedir}/swtpm_setup.log" \
                                --tpm "${workdir}/${libtpmsdir}/swtpm socket" \
                                ${profileopt:+${profileopt}}
```

mikkorapeli-linaro: IMO this is wrong. And caused me to spend a day figuring out that "path/to/swtpm socket" was the correct way to define custom swtpm path.

With swtpm is found from PATH, then " socket" is added. IMO it should be added with custom swtpm path too.

stefanberger: > With swtpm is found from PATH, then " socket" is added. IMO it should be added with custom swtpm path too.

I think running swtpm_setup with the `--tpm` parameter is a special case and in this case you should just provide the full path to the executable.

stefanberger: ... just run `$(type -P myswtpm)` on command line and it's done.

mikkorapeli-linaro: If the test is failing with this patch, then I can provide fxi for it. I presume something like:

```
--- a/tests/test_tpm2_libtpms_versions_profiles
+++ b/tests/test_tpm2_libtpms_versions_profiles
@@ -686,7 +686,7 @@ function swtpm_setup_create_profile_state()
                                --tpmstate "${statedir}" \
                                --config "${workdir}/swtpm_setup.conf" \
                                --log "${statedir}/swtpm_setup.log" \
-                               --tpm "${workdir}/${libtpmsdir}/swtpm socket" \
+                               --tpm "${workdir}/${libtpmsdir}/swtpm" \
                                ${profileopt:+${profileopt}}; then
                                echo "Error: Could not run swtpm_setup with libtpms from branch ${branch} with profile ${profile}"
                                err=1
```

Or is the "--tpm" argument used to provide other, non-swtpm TPM device paths too? I did not think this was possible.

mikkorapeli-linaro: Full path to "swtpm" with swtpm_setup --tpm argument was failing. The command line help for swtpm_setup only mentioned --tpm as path to "swtpm" binary, not "path to 'swtpm' binary and argument 'socket'".

mikkorapeli-linaro: > ... just run `$(type -P myswtpm)` on command line and it's done.

This does not work. There needs to be the "socket" argument too. Without it older versions of swtpm fail with "support for TPM2 not in swtpm binary".

stefanberger: This works:

```
swtpm_setup \
  --tpm "$(type -P swtpm) socket" \
  --tpm2 \
  --overwrite \
  --tpm-state .
```

To maintain backwards compatibility we cannot make that change anyway.

mikkorapeli-linaro: This was not at all obvious. I can add this to the command line help but I would really prefer to add the " socket" argument automatically.

Backwards compatibility is important but IMO it should not cover design bugs.

So really just fix the documentation then?

stefanberger: > So really just fix the documentation then?

Yes: https://github.com/stefanberger/swtpm/pull/958

mikkorapeli-linaro:
Fixed documentation instead in https://github.com/stefanberger/swtpm/pull/958

PR #958, opened 2024-12-10
man: Improve swtpm_setup --tpm option documentation

stefanberger:
Improve the swtpm_setup --tpm option documentation that did not mention that the socket option must be passed along when swtpm is being used.

### DIFF
--- a/man/man8/swtpm_setup.pod
+++ b/man/man8/swtpm_setup.pod
@@ -37,10 +37,18 @@ will be used.
 Path where the TPM's state will be written to; this is a mandatory argument.
 Prefix with dir:// to use directory backend, or file:// to use linear file.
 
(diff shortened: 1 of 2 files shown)


man/man8/swtpm_setup.pod
```diff
@@ -37,10 +37,18 @@ will be used.
 Path where the TPM's state will be written to; this is a mandatory argument.
 Prefix with dir:// to use directory backend, or file:// to use linear file.
 
-=item B<--tpm <path to executable>>
+=item B<--tpm "path-to-executable socket">
```
mikkorapeli-linaro, line 40: Could you add this to the swtpm_setup.c help text too?

mikkorapeli-linaro:
Thanks for the quick response, looks good to me.

PR #995, opened 2025-04-09
Fix SELinux related bugs in the swtpm.spec file

ajeeth-adithya-nutanix:
•  Uninstall ‘swptm_svirt’ selinux module first before uninstalling the ‘swtpm’ module.
•  Add the %selinux_relabel_pre macro to back up the original file contexts lists.

Signed-off-by: Ajeeth Adithya <ajeeth.adithya@nutanix.com>

### DIFF
--- a/swtpm.spec
+++ b/swtpm.spec
@@ -134,7 +134,7 @@ done
 
 %postun selinux
 if [ $1 -eq  0 ]; then
-  for p in swtpm swtpm_libvirt swtpm_svirt; do
+  for p in swtpm_svirt swtpm_libvirt swtpm; do
     %selinux_modules_uninstall -s %{selinuxtype} $p
   done
 fi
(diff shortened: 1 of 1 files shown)


stefanberger:
> • Uninstall ‘swptm_svirt’ selinux module first before uninstalling the ‘swtpm’ module. • Add the %selinux_relabel_pre macro to back up the original file contexts lists.

Thanks for the patch!
Were there any negative side-effects by not having it done in proper order? If so, can you describe them.

ajeeth-adithya-nutanix:
@stefanberger sorry I did not see your comment. Updated the commit message. 

The triggering of the %postun section did not cause any issues on its own. It's just that the swtpm module is never uninstalled.

stefanberger:
> @stefanberger sorry I did not see your comment. Updated the commit message.

Thanks.
> 
> The triggering of the %postun section did not cause any issues on its own. It's just that the swtpm module is never uninstalled.

Yes, you're right:
```
$ sudo rpm -e swtpm-selinux --nodeps
$ sudo semodule -n -s targeted -l | grep swtpm
swtpm
```

I will apply it also to the stable-0.9 and stable-0.10 branches and update the Fedora spec file.

stefanberger:
@elmarco FYI.

PR #1011, opened 2025-04-28
Fix some issues preventing the build on GNU/Hurd systems from succeeding

stefanberger:
The build of swtpm currently fails on GNU/Hurd systems due to some issues related to header files. Fix them.
However, the test cases require either netstat or ss tools, of which neither seems to be available. Therefore, `--disable-tests` needs to be passed to configure script. Also socat 1.7.x seems broken and cannot run the unix command socket tests.

### DIFF
--- a/include/swtpm/tpm_ioctl.h
+++ b/include/swtpm/tpm_ioctl.h
@@ -23,6 +23,10 @@
 #include <sys/ioccom.h>
 #endif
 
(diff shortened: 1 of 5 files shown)


stefanberger:
Something seems wrong with the socat version 1.7.4.4 available on GNU/Hurd when using it with source FILE: and destination UNIX-CLIENT:.

PR #1035, opened 2025-05-21
migrate to libfuse3 as fuse2 contains unaddressed security issue

leo9800:
migrate to libfuse3 as fuse2 contains unaddressed security issue (libfuse/libfuse#15)

 - configure.ac: update fuse package name
 - src/swtpm/cuse_tpm.c: migrate invocation of `fuse_session_loop_mt()`

with successful invocation of the below commands, it seems the patched source is working?

```
autoreconf --install --force
./configure --with-cuse
make
make check
```

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -200,10 +200,10 @@ AC_ARG_WITH([cuse],
 )
 
 if test "$with_cuse" != "no"; then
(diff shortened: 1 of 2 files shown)


stefanberger:
You should add a Signed-off-by: to the patch and prefix your commit with "cuse: migrate to libfuse3...". Otherwise LGTM.

PR #1036, opened 2025-05-22
cuse: add support to libfuse3 while avoid introducing breaking change

leo9800:
libfuse2 contains unaddressed security issue. (libfuse/libfuse#15)

this patch adds a configuration flag `--with-fuse3`, when it is explicitly set, swtpm would be built against libfuse3 instead of libfuse2.

while `--with-fuse3` is not set, the package is built against libfuse2 as conventional, this avoid breaking existing build systems from various distros.

	- configure.ac: add flag `--with-fuse3`, which requires enabling explicitly
	- src/swtpm/cuse_tpm.c: add libfuse3 support, mainly MARCOs and `fuse_session_loop_mt()` reimplementation

with successful building and testing (returns 0) with both 2 configurations below, it seems the patched source is working?

```
autoreconf --install --force
./configure --with-cuse  # fuse2
make clean && make && make check
```

```
autoreconf --install --force
./configure --with-cuse --with-fuse3  # fuse3
make clean && make && make check
```

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -200,24 +200,24 @@ AC_ARG_WITH([cuse],
 )
 
 if test "$with_cuse" != "no"; then
-    LIBFUSE_CFLAGS=$(pkg-config fuse --cflags 2>/dev/null)
-    if test $? -ne 0; then
-        if test "$with_cuse" = "yes"; then
-            AC_MSG_ERROR("Is fuse-devel installed? -- could not get cflags for libfuse")
-        else
-            with_cuse=no
-        fi
-    else
+    if pkg-config fuse3; then
+        LIBFUSE_CFLAGS=$(pkg-config fuse3 --cflags)
+        LIBFUSE_LIBS=$(pkg-config fuse3 --libs)
         with_cuse=yes
+    elif pkg-config fuse; then
+        LIBFUSE_CFLAGS="$(pkg-config fuse --cflags) -DWITH_FUSE2"
+        LIBFUSE_LIBS=$(pkg-config fuse --libs)
+        with_cuse=yes
+    else
+        if test "$with_cuse" = "yes"; then
+            AC_MSG_ERROR("Is fuse3-devel or fuse-devel installed? -- could not found libfuse3 or libfuse")
+        fi
+        with_cuse=no
     fi
 fi
 
 dnl with_cuse is now yes or no
-if test "$with_cuse" != "no"; then
-    LIBFUSE_LIBS=$(pkg-config fuse --libs)
-    if test $? -ne 0; then
-        AC_MSG_ERROR("Is fuse-devel installed? -- could not get libs for libfuse")
-    fi
+if test "$with_cuse" = "yes"; then
     AC_SUBST([LIBFUSE_CFLAGS])
     AC_SUBST([LIBFUSE_LIBS])
     AC_DEFINE_UNQUOTED([WITH_CUSE], 1,
@@ -229,6 +229,7 @@ if test "$with_cuse" != "no"; then
     fi
     AC_SUBST([GTHREAD_LIBS])
 fi
+
 AM_CONDITIONAL([WITH_CUSE],[test "$with_cuse" = "yes"])
 AC_MSG_RESULT($with_cuse)
 
@@ -675,6 +676,8 @@ echo "        GLIB_LIBS = $GLIB_LIBS"
 echo "      GNUTLS_LIBS = $GNUTLS_LIBS"
 echo "         GMP_LIBS = $GMP_LIBS"
 echo "       GMP_CFLAGS = $GMP_CFLAGS"
+echo "   LIBFUSE_CFLAGS = $LIBFUSE_CFLAGS"
+echo "     LIBFUSE_LIBS = $LIBFUSE_LIBS"
 echo
 echo "TSS_USER=$TSS_USER"
 echo "TSS_GROUP=$TSS_GROUP"
--- a/src/swtpm/cuse_tpm.c
+++ b/src/swtpm/cuse_tpm.c
@@ -54,7 +54,12 @@
 #include <arpa/inet.h>
 #include <dirent.h>
 
+#ifdef WITH_FUSE2
 #include <fuse/cuse_lowlevel.h>
+#else
+#define FUSE_USE_VERSION 35
+#include <fuse3/cuse_lowlevel.h>
+#endif
 
 #include <glib.h>
 
@@ -1581,7 +1586,11 @@ ptm_cuse_lowlevel_main(int argc, char *argv[], const struct cuse_info *ci,
         return 1;
 
     if (param->seccomp_action == SWTPM_SECCOMP_ACTION_NONE && mt)
+#ifdef WITH_FUSE2
         ret = fuse_session_loop_mt(ptm_fuse_session);
+#else
+        ret = fuse_session_loop_mt(ptm_fuse_session, NULL);
+#endif
     else
         ret = fuse_session_loop(ptm_fuse_session);
 


leo9800:
this patch supersedes #1035 as it avoids BC, while #1035 not.

configure.ac
```diff
@@ -198,23 +198,57 @@ AC_ARG_WITH([cuse],
...
+    LIBFUSE_LIBS=$(pkg-config fuse3 --libs 2>/dev/null)
+    if test $? -ne 0; then
+        AC_MSG_ERROR("Is fuse3-devel installed? -- could not get libs for libfuse3")
+    fi
+    AC_SUBST([LIBFUSE_CFLAGS])
+    AC_SUBST([LIBFUSE_LIBS])
+    AC_DEFINE_UNQUOTED([WITH_CUSE], 1,
+                       [whether to build with CUSE interface])
+
+    GTHREAD_LIBS=$(pkg-config --libs gthread-2.0)
```
stefanberger, line 239: Why do you need gthread-2.0?  libfuse3 seems to need pthread:
```
$ pkg-config fuse3 --libs
-lfuse3 -lpthread
```

configure.ac
```diff
@@ -198,23 +198,57 @@ AC_ARG_WITH([cuse],
...
 )
+AC_ARG_WITH([fuse3],
+            AS_HELP_STRING([--with-fuse3],[build with libfuse3 instead of libfuse2]),
+            [],
+            [with_fuse3=check]
+)
 
 if test "$with_cuse" != "no"; then
-    LIBFUSE_CFLAGS=$(pkg-config fuse --cflags 2>/dev/null)
-    if test $? -ne 0; then
```
stefanberger, line 204: It was checking whether fuse was installed and if it could not be found but the user requested it, it would throw the error. Otherwise it would build with cuse just if fuse was found. I think this behavior should remain, but now first test for fuse3 and if not available then fall back to fuse.

Another good thing would be to display the LIBFUSE_CFLAGS and LIBFUSE_LIBS at the bottom of the script so the user sees what it links against.

leo9800:
so such logic could be implemented with `configure.ac`?

1. if `--with-cuse` is not set, or `--without-cuse` is explicitly set, don't even bother checking any version of libfuse.
2. if `--with-cuse` is set, check availablity of fuse3 with `pkg-config`, if available, set cflags and ldflags
3. if fuse3 is not available, check availablity of fuse2
4. if fuse2 is neither available, fail with error message

stefanberger:
I think something along these lines is simplest:

```
if test "$with_cuse" != "no"; then
    FUSELIB=fuse3
    LIBFUSE_CFLAGS=$(pkg-config $FUSELIB --cflags 2>/dev/null)
    if test $? -ne 0; then
        FUSELIB=fuse
        LIBFUSE_CFLAGS=$(pkg-config $FUSELIB --cflags 2>/dev/null)
        if test $? -ne 0; then
            if test "$with_cuse" = "yes"; then
                AC_MSG_ERROR("Is fuse/fuse3-devel installed? -- could not get cflags for libfuse or libfuse3")
            else
                with_cuse=no
            fi
        else
            with_cuse=yes
        fi
    else
        with_cuse=yes
    fi
fi

dnl with_cuse is now yes or no
if test "$with_cuse" != "no"; then
    LIBFUSE_LIBS=$(pkg-config $FUSELIB --libs)
```

configure.ac
stefanberger: You are right. If we compile with the CUSE interface then gthread is needed by the CUSE swtpm for a threadpool.

configure.ac
```diff
@@ -198,23 +198,57 @@ AC_ARG_WITH([cuse],
...
-        fi
-    else
+    if test "$with_cuse" = "yes"; then
         with_cuse=yes
+    else
+        with_cuse=no
+    fi
+fi
+
+if test "$with_fuse3" != "no"; then
```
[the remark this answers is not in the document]
stefanberger: Once FUSELIB is set to either fuse or fuse3 you could see whether fuse3 was picked up...
```
if test "$with_fuse3" = "yes"; then
  if "$FUSELIB" != "fuse3"; then
    AC_MSG_ERROR()
  fi
fi
```

leo9800:
@stefanberger reimplemented fuse3->fuse2->failout logic 😆

also, `LIBFUSE_CFLAGS` and `LIBFUSE_LIBS` would be print upon `./configure`

test with both fuse3 & fuse2 passed in docker.

configure.ac
```diff
@@ -200,25 +200,22 @@ AC_ARG_WITH([cuse],
...
 
 dnl with_cuse is now yes or no
-if test "$with_cuse" != "no"; then
-    LIBFUSE_LIBS=$(pkg-config fuse --libs)
-    if test $? -ne 0; then
-        AC_MSG_ERROR("Is fuse-devel installed? -- could not get libs for libfuse")
-    fi
-    AC_SUBST([LIBFUSE_CFLAGS])
+if test "$with_cuse" == "yes"; then
+	AC_SUBST([LIBFUSE_CFLAGS])
```
stefanberger, line 218: Use a single '=' in comparisons in shell scripts.
Indentation of AC_SUBST with spaces.
And please squash the two patches into one. Keep the description and title of the first.

configure.ac
```diff
@@ -200,25 +200,22 @@ AC_ARG_WITH([cuse],
...
+		LIBFUSE_LIBS=$(pkg-config fuse3 --libs)
+		with_cuse=yes
+	elif pkg-config fuse; then
+		LIBFUSE_CFLAGS="$(pkg-config fuse --cflags) -DWITH_FUSE2"
+		LIBFUSE_LIBS=$(pkg-config fuse --libs)
+		with_cuse=yes
+	else
+		AC_MSG_ERROR("Is fuse3-devel or fuse-devel installed? -- could not found libfuse3 or libfuse")
+	fi
 fi
```
stefanberger, line 214: When none of them is installed, then this doesn't work anymore:

```
./configure --prefix=/usr
[...]
checking for clock_gettime in -lc... yes
checking for whether to build with CUSE interface... configure: error: "Is fuse3-devel or fuse-devel installed? -- could not found libfuse3 or libfuse"
```

leo9800: actually this is the expected behaviour, when user specified `--with-cuse` but neither fuse3 nor fuse2 is installed, `./configure` should fail and notify about it.

or should it modify `with_cuse` to `no` and then continue? maybe not, because the product it build does not contains `swtpm_cuse`, which is a mismatching to `--with-cuse` flag.

configure.ac
leo9800: ooops
a typo ...

configure.ac
stefanberger: I did not `specify --with-cuse` and in this case it previously was building with CUSE when it detected fuse , did not build CUSE otherwise but was throwing an error if it was request with `--with-cuse` but fuse could not be found. I just want the old behavior. My suggestion for when to error out with AC_MSG_ERROR should fix it.

configure.ac
```diff
@@ -200,25 +200,22 @@ AC_ARG_WITH([cuse],
...
+	if pkg-config fuse3; then
+		LIBFUSE_CFLAGS=$(pkg-config fuse3 --cflags)
+		LIBFUSE_LIBS=$(pkg-config fuse3 --libs)
+		with_cuse=yes
+	elif pkg-config fuse; then
+		LIBFUSE_CFLAGS="$(pkg-config fuse --cflags) -DWITH_FUSE2"
+		LIBFUSE_LIBS=$(pkg-config fuse --libs)
+		with_cuse=yes
+	else
+		AC_MSG_ERROR("Is fuse3-devel or fuse-devel installed? -- could not found libfuse3 or libfuse")
```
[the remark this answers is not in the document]
stefanberger: This suggestion should bring back the old behavior.

PR #1042, opened 2025-06-29
configure.ac: Use pkgconf macros, use m4sh, remove obsolete sections

salahcoronya:
This cleans up configure.ac, the hard coding of the pkgconf is causing cross compile problems. The non-m4sh can cause problems with pkgconf macros though. These patches fix that up.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -82,21 +82,21 @@ AS_IF([test "x$with_selinux" != "xno"],
 
 AM_CONDITIONAL([WITH_SELINUX], [test "x$with_selinux" = "xyes"])
 
-if test "$prefix" = "/usr" && test "$sysconfdir" = '${prefix}/etc'; then
-	sysconfdir="/etc"
-fi
-if test "$prefix" = "" && test "$datarootdir" = '${prefix}/share'; then
-	datarootdir="/usr/share"
-fi
-if test "$prefix" = "/usr" && test "$localstatedir" = '${prefix}/var'; then
-	localstatedir="/var"
-fi
-if test "x$prefix" = "xNONE"; then
-	prefix="/usr/local"
-fi
-if test "x$exec_prefix" = "xNONE"; then
-	exec_prefix=$prefix
-fi
+AS_IF([test "$prefix" = "/usr" && test "$sysconfdir" = '${prefix}/etc'],
+    [sysconfdir="/etc"]
+)
+AS_IF([test "$prefix" = "" && test "$datarootdir" = '${prefix}/share'],
+    [datarootdir="/usr/share"]
+)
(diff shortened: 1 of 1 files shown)


stefanberger:
That's a lot of changes. I will not take them all at once but would like to cherry-pick them from your PR here into another PR and make this list here shorter over the next few days/weeks. Is this OK with you?

stefanberger:
... and thanks a lot for the PR!

salahcoronya:
That's fine, feel free to squash/rearrange as needed

stefanberger:
Could you rebase this series on master~9, then master~8 ... master~1, master, and push again the remaining ones. One of the patches doesn't easily rebase due to a whitespace fix I did after a `[yes]`.

configure.ac
```diff
@@ -52,31 +52,6 @@ dnl Allow 64-bit file API on 32-bit systems. Without the change even small
 dnl files will fail to stat any files on filesystems with 64-bit inodes.
 AC_SYS_LARGEFILE
 
-DEBUG=""
-AC_MSG_CHECKING([for debug-enabled build])
-AC_ARG_ENABLE(debug, AS_HELP_STRING([--enable-debug],[create a debug build]),
```
stefanberger, line 57: This is the last patch that I did not apply so far.

I agree that we can remove this part here.

configure.ac
```diff
@@ -117,30 +92,6 @@ if test "$with_selinux" != "no"; then
...
-	exec_prefix=$prefix
-fi
-SYSCONFDIR=`eval echo $sysconfdir`
-DATAROOTDIR=`eval echo $datarootdir`
-LOCALSTATEDIR=`eval echo $localstatedir`
-BINDIR=`eval echo $bindir`
-AC_SUBST([SYSCONFDIR])
-AC_SUBST([DATAROOTDIR])
-AC_SUBST([LOCALSTATEDIR])
-AC_SUBST([BINDIR])
```
stefanberger, line 142: These here are necessary for conversion of src/utils/swtpm_conf.h.in and sample/swtpm-localca.conf.in.
If I remove all the `if test` statements above it I get this here in src/utils/swtpm_conf.h:

```
#define SYSCONFDIR "/etc"
#define BINDIR "NONE/bin"
#define DATAROOTDIR "/usr/share"
```

It seems to be that the NONE also needs to be tested for with these test statements. I don't think we can remove this block here.

salahcoronya: Your right, its still needed. So I split this patch into 2: One to remove the debug checking CFLAGS, and another to m4shify that section

stefanberger:
All merged. Thanks.

PR #1064, opened 2025-10-01
WIP: swtpm_cert: Enable CAs with ML-DSA key to sign a certifcate

stefanberger:
Test for GNUTLS_PK_MLDSA44 to detect whether GnuTLS supports ML-DSA.

Only SHAKE-256 can be used for hashing when ML-DSA is used for signing:

https://github.com/gnutls/gnutls/blob/df24a53136f188d77aaffe66316b0fb6ba720d40/lib/algorithms/sign.c#L405-L428

The problem is now that the size of NVRAM indices is limited to MAX_NV_INDEX_SIZE = 2048, which is too small for a certificate created even with ML-DSA-44, which is around 2757 bytes long.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -220,6 +220,25 @@ AS_IF([test "x$with_gnutls" != "xno"],
         [with_gnutls=no]
     )
     AS_IF([test "x$with_gnutls" != "xno"],[with_gnutls="yes"])
+    AS_IF([test "x$with_gnutls" = "xyes"],
+          [AC_COMPILE_IFELSE(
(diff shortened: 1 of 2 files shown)


stefanberger:
Writing the larger certificates into NVRAM indices will only work with
- this PR applied to libtpms: https://github.com/stefanberger/libtpms/pull/535
- default-v2 profile

stefanberger:
This was a GnuTLS based patch. This patch is not needed anymore after migration to OpenSSL since it is now also supported in PR #1099

PR #1075, opened 2026-02-23
Implement support for capturing TPM packet exchanges to pcap files

stefanberger:
This PR extends swtpm with support for capturing TPM packet exchanges and writing then into a pcapng formatted file.

The pcapng format is described here: https://www.ietf.org/archive/id/draft-tuexen-opsawg-pcapng-03.html
And also here: https://pcapng.com/

### DIFF
--- a/tests/Makefile.am
+++ b/tests/Makefile.am
@@ -65,6 +65,7 @@ TESTS += \
 	test_tpm2_hashing3 \
 	test_tpm2_migration_key \
 	test_tpm2_partial_reads \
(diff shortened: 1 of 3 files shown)


henryreed:
@stefanberger : I am very sorry I haven't been able to look at this until now--between preparing to give a conference workshop and other work, my free time was significantly constrained. This is fantastic, however! Thanks for making this happen. Really excited to see this make it into Proxmox and other tooling later down the line.

PR #823, opened 2023-08-21
Add support for creating IAK and IDevID keys and certificates

stefanberger:
This PR extends the tools swtpm_cert, swtpm_localca, and swtpm_setup to create IAK and IDevID keys and certificates.

### ToDo:
- [ ] swtpm_setup: Check NVRAM area flags
- [ ] swtpm_setup: NVRAM indices currently used are wrong. Which ones to use?  We need to wait for an update of the specs that show the NVRAM indices for the certs.
- [X] swtpm_setup: Create serialNumber for subject from `<TCG Manuf Code>:<EK Authority Key Id>:<Ek Cert Serial Number>`
- [ ] swtpm_cert: Is the ASN.1 in the certs ok? Properly nested?
- [X] man pages
- [X] test cases

PR #1109, opened 2026-04-21
Add support for creating IAK and IDevID keys and certificates (v2)

stefanberger:
This is a rebase of the code in PR #823.

This PR extends the tools swtpm_cert, swtpm_localca, and swtpm_setup to create IAK and IDevID keys and certificates.

### ToDo:
- [X] swtpm_setup: Check NVRAM area flags
- [X] swtpm_setup: NVRAM indices currently used are wrong. Which ones to use?  We need to wait for an update of the specs that show the NVRAM indices for the certs.
- [X] swtpm_setup: Create serialNumber for subject from `<TCG Manuf Code>:<EK Authority Key Id>:<Ek Cert Serial Number>`
- [X] swtpm_cert: Is the ASN.1 in the certs ok? Properly nested?
- [X] man pages
- [X] test cases

### DIFF
--- a/man/man8/swtpm_setup.pod
+++ b/man/man8/swtpm_setup.pod
@@ -113,6 +113,8 @@ I<ek2keyalgo> entry in the I<swtpm_setup.conf> configuration file.
 =item B<--iakkeyalgo <key algorithm>> (since v0.11)
 
 Choose the key algorithm for the IAK key that can be used for signing.
+An IAK requires an EK certificate to be created to refer to in its own
+certificate.
 
 Valid values are: ecc_nist_p256 or secp256r1, ecc_nist_p384 or secp384r1,
 ecc_nist_p521 or secp521r1, rsa2048, rsa3072, rsa4096.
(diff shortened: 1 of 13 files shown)


man/man8/swtpm_setup.pod
```diff
@@ -367,6 +368,12 @@ The I<--create-config-files> option is supported.
 The I<--reconfigure> option is supported and allows the reconfiguration of
 the active PCR banks.
 
+=item B<cmdarg-no-iak> (since v0.11)
```
stefanberger, line 372: Offer `--aikkeyalgo` and `--idevidkeyalgo` that user must provide to get the keys with a given algorithm

src/swtpm_setup/swtpm_setup.h
```diff
@@ -14,4 +14,7 @@
 
 extern gchar *gl_LOGFILE;
 
+#define TPM_CAP_TPM_PROPERTIES 6
+#define TPM_PT_MANUFACTURER 0x105
```
stefanberger: `TPM_` -> `TPM2_`

PR #1121, opened 2026-05-07
Query for the buffer size of the TPM 2 in libtpms v0.11

stefanberger:
Introduce a hard dependency on libtpms v0.11, which will be needed for PQC support. Query the TPM 2's buffer size using the new TPMPROP_TPM2_BUFFER_MAX introduced in libtpms v.0.11. For now it's 4kb but will become 8kb once PQC enablement is there.


baloo:
Is there a branch for libtpms 0.11 already?
I can't find a release or even a branch.

EDIT: nevermind, that's just master. That bumped 2 years ago.

PR #1122, opened 2026-05-07
Sync permanent state blob upon GET_STATEBLOB cmd

adrobvz:
We've seen a live migration failure in production, with Win2k25 guest, with the following symptoms:

QEMU dest log:
```
qemu-kvm: tpm-emulator: Setting the stateblob (type 2) failed with a TPM error 0x800
qemu-kvm: error while loading state for instance 0x0 of device 'tpm-emulator'
qemu-kvm: load of migration failed: Input/output error
```

swtpm dest log:
```
main: Initializing TPM at Fri May  1 14:04:10 2026                              
mainLoop:                                                                       
SWTPM_NVRAM_Validate_Dir: Rooted state path /path/to/shared/storage             
SWTPM_NVRAM_Validate_Dir: Rooted state path /path/to/shared/storage             
SWTPM_NVRAM_LoadData: No such file /path/to/shared/storage/tpm2-00.permall                                                                                     
Data client disconnected 
```

The root cause is not perfectly clear, either orchestration layer left the blob out of sync, or some kind of storage error occured when the blob was initially written on the src node.  Either way, I suggest an additional sync for the permanent state blob as a potential workaround for such errors, similar to how it's done with volatile state.

### DIFF
--- a/src/swtpm/swtpm_nvstore.c
+++ b/src/swtpm/swtpm_nvstore.c
@@ -481,6 +481,37 @@ TPM_RESULT SWTPM_NVRAM_Store_Volatile(void)
     return rc;
 }
 
+/*
+ * Flush libtpms's in-memory PERMANENT state to disk so that the next
+ * SWTPM_NVRAM_GetStateBlob(PERMANENT) returns a blob consistent with what
+ * libtpms is actually running. Counterpart to SWTPM_NVRAM_Store_Volatile().
+ *
+ * Without this, GET_STATEBLOB(PERMANENT) on the migration-out path reads a
+ * file that may be missing, empty, or stale relative to libtpms RAM (the
+ * file is only written by _plat__NvCommit() on UT_NV-class commands), and
+ * an empty/short-read blob then poisons the destination's state restore.
+ */
+TPM_RESULT SWTPM_NVRAM_Store_Permanent(void)
+{
+    TPM_RESULT     rc = 0;
+    char           *name = TPM_PERMANENT_ALL_NAME;
+    uint32_t       tpm_number = 0;
+    unsigned char  *buffer = NULL;
+    uint32_t       buflen = 0;
+
+    TPM_DEBUG(" SWTPM_Store_Permanent: Name %s\n", name);
+    if (rc == 0) {
+        rc = TPMLIB_GetState(TPMLIB_STATE_PERMANENT, &buffer, &buflen);
+    }
+    if (rc == 0 && buflen > 0) {
+        rc = SWTPM_NVRAM_StoreData(buffer, buflen, tpm_number, name);
+    }
+
+    free(buffer);
+
+    return rc;
+}
+
 static TPM_RESULT
 SWTPM_NVRAM_KeyParamCheck(uint32_t keylen,
                           enum encryption_mode encmode)
--- a/src/swtpm/ctrlchannel.c
+++ b/src/swtpm/ctrlchannel.c
@@ -175,6 +175,9 @@ static int ctrlchannel_return_state(ptm_getstate *pgs, int fd,
     if (!blobname)
         res = TPM_FAIL;
 
+    if (res == 0 && blobtype == PTM_BLOB_TYPE_PERMANENT)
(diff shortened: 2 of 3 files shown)


dlunev:
This state could be easily achieved if we will migrate Win2025 just being installed, f.e. at some installer stage when TPM is not yet initialized for the very first time.

stefanberger:
> We've seen a live migration failure in production, with Win2k25 guest, with the following symptoms:

A migration error should not occur due to the VM that is running, like Win2k25. Have you had such issues with any other workloads or do they all work fine? Does it always happen with Win2k25 or is it 'a live migration failure' that only occurred once or once in a while?

> 
> QEMU dest log:
> 
> ```
> qemu-kvm: tpm-emulator: Setting the stateblob (type 2) failed with a TPM error 0x800
> qemu-kvm: error while loading state for instance 0x0 of device 'tpm-emulator'
> qemu-kvm: load of migration failed: Input/output error
> ```
> 
> swtpm dest log:
> 
> ```
> main: Initializing TPM at Fri May  1 14:04:10 2026                              
> mainLoop:                                                                       
> SWTPM_NVRAM_Validate_Dir: Rooted state path /path/to/shared/storage             
> SWTPM_NVRAM_Validate_Dir: Rooted state path /path/to/shared/storage             
> SWTPM_NVRAM_LoadData: No such file /path/to/shared/storage/tpm2-00.permall                                                                                     
> Data client disconnected 
> ```
> 
> The root cause is not perfectly clear, either orchestration layer left the blob out of sync, or some kind of storage error occured when the blob was initially written on the src node. Either way, I suggest an additional sync for the permanent state 

To me it seems that it was an issue with the storage.

> blob as a potential workaround for such errors, similar to how it's done with volatile state.

src/swtpm/swtpm_nvstore.c
```diff
@@ -481,6 +481,37 @@ TPM_RESULT SWTPM_NVRAM_Store_Volatile(void)
     return rc;
 }
 
+/*
+ * Flush libtpms's in-memory PERMANENT state to disk so that the next
+ * SWTPM_NVRAM_GetStateBlob(PERMANENT) returns a blob consistent with what
+ * libtpms is actually running. Counterpart to SWTPM_NVRAM_Store_Volatile().
```
stefanberger, line 487: This makes sense and it shouldn't affect anything negatively. However, I was under the impression that the PERMANENT state was always stored whenever a command 'affected' it, e.g. when an NVRAM space was created or undefined etc., and therefore the latest state was always on disk. So I am a bit surprised that this is necessary and that nobody else complained about it so far.

src/swtpm/swtpm_nvstore.c
```diff
@@ -481,6 +481,37 @@ TPM_RESULT SWTPM_NVRAM_Store_Volatile(void)
...
+ *
+ * Without this, GET_STATEBLOB(PERMANENT) on the migration-out path reads a
+ * file that may be missing, empty, or stale relative to libtpms RAM (the
```
stefanberger, line 490: You don't happen to have a reproducer for how we could generate a 'missing' or 'empty file' situation other than using `rm` intentionally, do you?

adrobvz:
Hello Stefan,

I spent some more time trying to reproduce the issue by hand. We thought migrating a VM before the guest OS is able to send any TPM related commands might be it. I tried Win11 during installation, then empty VM booted into UEFI shell - no luck. The reason is because the state is being created before VM is even launched by `swtpm_setup`, which is in turn launched by libvirt.

So I still don't have a solid reproduce for this, except for doing manual `rm` / `truncate`. However, as you mentioned above, the situation when there's not the latest state on the disk shouldn't occur. Shouldn't we consider having that additional sync? Having it does help with the naive `truncate` reproduction.

stefanberger:
>  Shouldn't we consider having that additional sync? Having it does help with the naive truncate reproduction.

Thanks for running some more tests.

My point is that it shouldn't be necessary because whenever a command is sent to the TPM 2 that alters its permanent state then the permanent state should be written to disk automatically. If this wasn't the case then we would have to have it write its state to disk every time we shut it down as well...

PR #1141, opened 2026-08-10
Changes required to run on OpenBSD with LibreSSL

moritzbuhl:
This is all the changes I needed to create a (working?) port on OpenBSD.
The recent PQC changes are not yet working with LibreSSL so I had to work around them.
I purposefully did not indent the `! LibreSSL` branches to not clutter the diffs.

If it makes sense to upstream parts of these changes or if these changes do not make sense at all, I appreciate any feedback.
For now, I just wanted to put these changes somewhere.

I haven't properly tested the binary itself because I have a little bit of a bootstrapping problem with TPMs and OpenBSD.
The test log is here:
https://gist.github.com/moritzbuhl/9ef2d0e89c15e4f73532ff90449f1003

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -115,7 +115,7 @@ AC_ARG_WITH([openssl],
             [])
 
 AS_CASE([$cryptolib],[openssl],
-    [PKG_CHECK_MODULES([LIBCRYPTO],[libcrypto >= 3.5])
+    [PKG_CHECK_MODULES([LIBCRYPTO],[libcrypto >= 0.0])
     AC_CHECK_HEADERS([openssl/obj_mac.h],[],
         AC_MSG_ERROR(Is openssl-devel/libssl-dev installed?))
     AC_MSG_RESULT([Building with openssl crypto library])
@@ -139,7 +139,7 @@ AS_IF([test "$OPENSSL_CLITOOL" != "yes"],
 
 PKG_CHECK_MODULES([LIBTASN1],[libtasn1])
 
-PKG_CHECK_MODULES([LIBTPMS],[libtpms >= 0.11])
+PKG_CHECK_MODULES([LIBTPMS],[libtpms >= 0.10])
 
 
 AC_CHECK_LIB(c, clock_gettime, LIBRT_LIBS="", LIBRT_LIBS="-lrt")
--- a/src/swtpm/pcap.c
+++ b/src/swtpm/pcap.c
@@ -11,10 +11,10 @@
 #include <stdint.h>
 #include <time.h>
 #include <unistd.h>
+#include <arpa/inet.h>
 #include <netinet/ip.h>
 #include <netinet/tcp.h>
 #include <sys/uio.h>
-#include <arpa/inet.h>
 
 #include <glib.h>
 
@@ -74,6 +74,10 @@ struct packet {
     struct ip     iphdr; // Cygwin has 'struct ip' but not 'struct iphdr'
     struct tcphdr tcphdr;
 } __attribute__((packed));
+#define PACKET_ETHHDR_OFF	(sizeof(struct enhanced_packet_block_hdr))
+#define PACKET_IPHDR_OFF	(PACKET_ETHHDR_OFF + sizeof(struct ethhdr))
+#define PACKET_TCPHDR_OFF	(PACKET_IPHDR_OFF + sizeof(struct ip))
+#define PACKET_DATA_OFF		(PACKET_TCPHDR_OFF + sizeof(struct tcphdr))
 
 static __attribute__((noinline)) uint32_t calc_checksum(void *array, size_t array_len)
 {
@@ -93,21 +97,31 @@ static __attribute__((noinline)) uint32_t calc_checksum(void *array, size_t arra
      return check;
 }
 
-static void calc_ip_checksum(struct packet *packet)
+static void calc_ip_checksum(char *packet)
 {
+    struct ip iphdr;
+    struct ip *packet_iphdr = (struct ip *)&packet[PACKET_IPHDR_OFF];
     uint32_t check = 0;
 
-    packet->iphdr.ip_sum = 0;
+    memcpy(&iphdr, packet_iphdr, sizeof(iphdr));
+    iphdr.ip_sum = 0;
 
-    check = calc_checksum(&packet->iphdr, sizeof(packet->iphdr));
+    check = calc_checksum(&iphdr, sizeof(iphdr));
     check = (check & 0xffff) + (check >> 16);
+    check = htons(~check);
 
-    packet->iphdr.ip_sum = htons(~check);
+    memcpy(&packet_iphdr->ip_sum, &check, sizeof(check));
 }
 
-static void calc_tcp_checksum(struct packet *packet,
+static void calc_tcp_checksum(char *packet,
                               void *payload, size_t payload_len)
 {
+    struct ip iphdr;
+    struct tcphdr tcphdr;
+    struct ip *packet_iphdr = (struct ip *)&packet[PACKET_IPHDR_OFF];
+    struct tcphdr *packet_tcphdr = (struct tcphdr *)&packet[PACKET_TCPHDR_OFF];
+    memcpy(&iphdr, packet_iphdr, sizeof(iphdr));
+    memcpy(&tcphdr, packet_tcphdr, sizeof(tcphdr));
     struct {
         uint32_t saddr;
         uint32_t daddr;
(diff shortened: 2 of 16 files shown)


tests/test_tpm2_swtpm_setup_create_cert
```diff
@@ -143,7 +143,7 @@ for keysize in ${keysizes}; do
 		openssl x509 -inform der -in "${certfile}" -noout -text
 		exit 1
 	fi
-	if ! openssl x509 -inform der -in "${certfile}" -noout -text | grep "Key Usage" -A1 | \
+	if ! openssl x509 -inform der -in "${certfile}" -noout -text | grep -A1 "Key Usage" | \
```
stefanberger, line 146: I don't mind this patch. If it helps you, you can split it off and send a separate PR. The title has a typo. Swtich -> Switch and the patch needs a Signed-off-by.

tests/test_tpm2_swtpm_localca_pkcs11.test
```diff
@@ -76,6 +76,51 @@ fi
...
+		echo "Could not get public key for pkcs11 uri key ($pkcs11uri}."
+		echo "${msg}"
+		exit 1
+	fi
+
+	cat > "${template}" << EOF
+cn = "swtpm-localca"
+signing_key = "${pkcs11uri}"
+EOF
+	msg=$(certtool --generate-request --template "${template}" \
```
stefanberger, line 95: I got rid of gnutls entirely. So no more certtool, please.

tests/test_tpm2_swtpm_localca
```diff
@@ -115,13 +115,30 @@ do
...
+      exit 1
+    fi
+    if ! openssl verify \
+           -CAfile "${workdir}/swtpm-localca-rootca-cert.pem" \
+           -untrusted "${ISSUERCERT}" \
+         "${workdir}/ek.cert.pem"; then
+      echo "Error: Could not verify certificate chain."
+      exit 1
+    fi
+  else # no LibreSSL
```
stefanberger, line 133: this should probably become a `elif ! openssl verify` ... and you can then remove the `fi` below.

Now this patch only makes sense if we can support LibreSSL all the way through...

src/swtpm/pcap.c
```diff
@@ -278,33 +301,40 @@ static int pcap_file_write_idb(struct pcap_state *ps)
 /* Write a simulated TCP SYN/SYN+ACK/ACK or FIN/FIN+ACK/ACK exchange */
 static int pcap_file_tcp_flags(struct pcap_state *ps, uint8_t flag)
 {
-    struct packet packet;
-
-    pcap_packet_fill(&packet, true, 0, 0, ps);
-    packet.tcphdr.th_flags = flag;
+    char buf[PACKET_TCPHDR_OFF];
```
moritzbuhl, line 304: This is too small, isn't it?

src/swtpm/pcap.c
```diff
@@ -93,48 +97,59 @@ static __attribute__((noinline)) uint32_t calc_checksum(void *array, size_t arra
      return check;
 }
 
-static void calc_ip_checksum(struct packet *packet)
+static void calc_ip_checksum(char *packet)
 {
+    struct ip iphdr;
+    struct ip *packet_iphdr = (struct ip *)&packet[PACKET_IPHDR_OFF];
```
stefanberger, line 103: Why are these changes necessary? Does it cause crashes on some architectures?

tests/test_tpm2_swtpm_localca
stefanberger: The problem is that libtpms will require ML-DSA and ML-KEM (and possibly EdDSA) support and I am not sure what problems that will cause if LibreSSL were not to support those. Does it support them?

src/swtpm/pcap.c
moritzbuhl: First:
<pre>
In file included from /usr/ports/pobj/swtpm-0.10.2pre20260715/swtpm-db2386b2c1530de3a7fa57a8214c3630e785ec2f/src/swtpm/pcap.c:14:
/usr/include/netinet/ip.h:67:19: error: field has incomplete type 'struct in_addr'
   67 |         struct    in_addr ip_src, ip_dst; /* source and dest address */
      |                           ^
/usr/include/netinet/ip.h:67:11: note: forward declaration of 'struct in_addr'
   67 |         struct    in_addr ip_src, ip_dst; /* source and dest address */
      |                   ^
/usr/include/netinet/ip.h:67:27: error: field has incomplete type 'struct in_addr'
   67 |         struct    in_addr ip_src, ip_dst; /* source and dest address */
      |                                   ^
/usr/include/netinet/ip.h:67:11: note: forward declaration of 'struct in_addr'
   67 |         struct    in_addr ip_src, ip_dst; /* source and dest address */
      |                   ^
/usr/include/netinet/ip.h:185:19: error: field has incomplete type 'struct in_addr'
  185 |                         struct in_addr ipt_addr;
      |                                        ^
/usr/include/netinet/ip.h:67:11: note: forward declaration of 'struct in_addr'
   67 |         struct    in_addr ip_src, ip_dst; /* source and dest address */
      |                   ^
3 errors generated.
</pre>

Then, after only rearranging the headers:

<pre>
/usr/ports/pobj/swtpm-0.10.2pre20260715/swtpm-db2386b2c1530de3a7fa57a8214c3630e785ec2f/src/swtpm/pcap.c:74:19: error: field iphdr within 'struct packet' is less aligned than 'struct ip' and is usually due to
      'struct packet' being packed, which can lead to unaligned accesses [-Werror,-Wunaligned-access]
   74 |     struct ip     iphdr; // Cygwin has 'struct ip' but not 'struct iphdr'
      |                   ^
/usr/ports/pobj/swtpm-0.10.2pre20260715/swtpm-db2386b2c1530de3a7fa57a8214c3630e785ec2f/src/swtpm/pcap.c:75:19: error: field tcphdr within 'struct packet' is less aligned than 'struct tcphdr' and is usually due to
      'struct packet' being packed, which can lead to unaligned accesses [-Werror,-Wunaligned-access]
   75 |     struct tcphdr tcphdr;
      |                   ^
2 errors generated.
</pre>

And then (IIRC) I looked into the tcphdr struct and it is not representing the on-wire binary format to my understanding.

src/swtpm/pcap.c
```diff
@@ -93,48 +97,59 @@ static __attribute__((noinline)) uint32_t calc_checksum(void *array, size_t arra
...
 
-    packet->iphdr.ip_sum = htons(~check);
+    memcpy(&packet_iphdr->ip_sum, &check, sizeof(check));
 }
 
-static void calc_tcp_checksum(struct packet *packet,
+static void calc_tcp_checksum(char *packet,
                               void *payload, size_t payload_len)
 {
+    struct ip iphdr;
```
stefanberger, line 119: When packet is filled inside a function to which `struct packet *packet` is passed then the compiler knows that `packet` is `packed` and knows how to fill it out knowing that some alignments of 32bit values could be off. What could be a problem is when `&packet->ethhdr` or `&packet->iphdr` or `&packet->tcphdr` are passed to a function where these headers are access since then they are assumed to be aligned and could cause issues when >=32bit values are written to them (unlikely issue with 16bit). I don't see any such a function. `calc_checksum(&packet->iphdr, sizeof(packet->iphdr));` is NOT a problem since `calc_checksum` accesses the iphdr as a byte array, so no alignment is needed. In my opinion none of these packet-related changes are necessary.

moritzbuhl: Appending `__attribute__((packed))` to `struct ip` and `struct tcphdr` in `/usr/include/netinet/ip.h` and `/usr/include/netinet/tcp.h` makes the compiler happy, no need for these changes then.

stefanberger: I think these structures are naturally packed and do not need the explicitly `__attribute__((packed))`. They map to byte arrays that can directly go onto the wire. Also, wireshark does understand the pcap files that are created by swtpm and would complain if these structures were off.

configure.ac
```diff
@@ -139,7 +139,7 @@ AS_IF([test "$OPENSSL_CLITOOL" != "yes"],
 
 PKG_CHECK_MODULES([LIBTASN1],[libtasn1])
 
-PKG_CHECK_MODULES([LIBTPMS],[libtpms >= 0.11])
+PKG_CHECK_MODULES([LIBTPMS],[libtpms >= 0.10])
```
stefanberger, line 142: swtpm v0.11 will depend on libtpms being able to provide ML-DSA and ML-KEM.

src/swtpm_cert/ek-cert.c
```diff
@@ -176,11 +184,15 @@ typedef struct TCG_PCCLIENT_STORED_FULL_CERT_HEADER {
 
 static bool uses_hashless_signing(const EVP_PKEY *key)
 {
+#ifndef LIBRESSL_VERSION_NUMBER
```
stefanberger, line 187: My general comment is that these changes are a 'quite a bit' intrusive. My guess is other code bases will have the same issue once they support more recent OpenSSL APIs and should also run with LibSSL...

PR #1146, opened 2026-08-17

stefanberger:

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation of generated filenames and paths.
  * Prevented truncated or non-null-terminated values from being accepted when they exactly fill the available buffer.


stefanberger:
@coderabbitai review

PR #1147, opened 2026-08-17

stefanberger:

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved filename and filepath validation to prevent buffer overflows.
  * Ensured generated paths remain properly terminated for safer file handling.


stefanberger:
@coderabbitai review

PR #1148, opened 2026-08-17

stefanberger:

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation of generated filenames and file paths.
  * Prevented truncated output by ensuring space remains for the terminating character.


stefanberger:
@coderabbitai review

PR #1165, opened 2026-08-19
tests: check for TPM 2.0 in test_tpm2_ctrlchannel3

xpahos:
The test runs swtpm with --tpm2 but checks for TPM 1.2 support, causing it to be skipped by TPM 2-only builds. Check for TPM 2.0 instead.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected the test environment check so TPM 2.0-specific tests are skipped when the required TPM 2.0 emulator is unavailable.


stefanberger:
thanks!

PR #1176, opened 2026-09-03
cuse_tpm: validate --tpmstate backend-uri before assuming dir:// prefix

### LINKED ISSUE #1175: swtpm_cuse: --tpmstate backend-uri shorter than "dir://" advances pointer by 6 without prefix/length check -> OOB read in opendir()
## Summary

`tpm_start()` in the `swtpm_cuse` frontend unconditionally computes
`uri + strlen("dir://")` from the command-line-provided
`--tpmstate backend-uri=...` value. A URI shorter than 6 bytes produces a
pointer past the end of the heap allocation, which is then passed to
`opendir()` (and can reach `mkdir()`), reading out of bounds while
scanning for a NUL terminator.

## Affected version

swtpm 0.11.0 @ `db2386b2c1530de3a7fa57a8214c3630e785ec2f`; still present
on current master as of 2026-08-31 (`cuse_tpm.c:547`).

## Affected code

`src/swtpm/cuse_tpm.c:542-561` (option flow via `common.c:794-812`,
`tpmstate.c:66-74`):

```c
const char *uri = tpmstate_get_backend_uri();
const char *tpmdir = uri + strlen("dir://");   /* no prefix/length check */
...
dir = opendir(tpmdir);                          /* OOB read */
```

## Reproduction

```
swtpm_cuse ... --tpmstate backend-uri=x --flags not-need-init
```

(`not-need-init` makes the startup path run immediately.) The 2-byte
heap string `x\0` is advanced by 6; `opendir()` reads at least 1 byte out
of bounds. Harness attached (`swtpm-pocs-20260903.tar.gz`,
`candidate-0032/`):

```
SUMMARY: AddressSanitizer: heap-buffer-overflow cuse_tpm.c:551 in tpm_start
```

## Impact

Small fixed-offset heap OOB read at startup; crash under ASan, limited
practical impact with production allocators since the URI comes from the
local command line. Low severity, but trivial to fix.

## Suggested fix

Before the pointer advance, require `uri != NULL`,
`strlen(uri) >= strlen("dir://")`, and
`strncmp(uri, "dir://", 6) == 0`; fail startup with a clear error
otherwise. (Longer term, `swtpm_cuse` should accept the same backend URI
scheme set as the socket frontend instead of assuming `dir://`.)

---
Found by an independent vTPM security audit. Contact: Leyao, ICT CAS - [lileyao2002@gmail.com]

[swtpm-pocs-20260903.tar.gz](https://github.com/user-attachments/files/31775868/swtpm-pocs-20260903.tar.gz)

Johnsonajibi:
## Summary

Fixes #1175. `tpm_start()` in `src/swtpm/cuse_tpm.c` unconditionally computed `uri + strlen("dir://")` from the command-line-provided `--tpmstate backend-uri=...` value, with no check that `uri` actually starts with `"dir://"` or is even long enough:

```c
const char *uri = tpmstate_get_backend_uri();
const char *tpmdir = uri + strlen("dir://");   /* no prefix/length check */
...
dir = opendir(tpmdir);                          /* OOB read */
```

A URI shorter than 6 bytes (e.g. `x`) produces a pointer past the end of the string's allocation, which is then passed to `opendir()` (and can reach `mkdir()`), reading out of bounds while scanning for a NUL terminator.

## Fix

`swtpm_cuse` only supports the `dir://` backend, so require the prefix (and a non-NULL `uri`) before advancing past it, failing startup with a clear error otherwise — matching how the socket frontend (`swtpm_nvstore.c`'s `SWTPM_NVRAM_Init`) already validates `backend_uri` against known prefixes before dispatching to a backend.

## Verification

Compiles clean — both `swtpm` and `swtpm_cuse` (which share this translation unit) rebuild without warnings under the project's strict flags.

I wasn't able to exercise this live end-to-end in my environment — `swtpm_cuse` requires the CUSE kernel module, which isn't available there — so this is a code-level fix verified by inspection and compilation rather than a live-triggered PoC, unlike the other fixes in this campaign. Happy to iterate if testing surfaces anything.

---
> Found by an independent audit (credit: Leyao, ICT CAS). Fixed with assistance from Claude Code (claude.ai/code).

## Summary by CodeRabbit

- **Bug Fixes**
  - Added validation for TPM state backend URIs.
  - Invalid or missing directory-based URIs are now rejected with an error instead of potentially causing unsafe memory access.
  - TPM startup now fails safely when the configured backend URI is missing or uses an unsupported format.

### DIFF
--- a/src/swtpm/cuse_tpm.c
+++ b/src/swtpm/cuse_tpm.c
@@ -544,9 +544,18 @@ static int tpm_start(uint32_t flags, TPMLIB_TPMVersion l_tpmversion,
 {
     DIR *dir;
     const char *uri = tpmstate_get_backend_uri();
-    const char *tpmdir = uri + strlen("dir://");
+    const char *tpmdir;
 
     *res = TPM_FAIL;
+
+    /* uri is untrusted; must have the "dir://" prefix before advancing past it */
+    if (uri == NULL || strncmp(uri, "dir://", strlen("dir://")) != 0) {
+        logprintf(STDERR_FILENO,
+                  "Error: swtpm_cuse only supports the dir:// backend URI, "
+                  "got: %s\n", uri ? uri : "(null)");
+        return -1;
+    }
+    tpmdir = uri + strlen("dir://");
 
     dir = opendir(tpmdir);
     if (dir) {


src/swtpm/cuse_tpm.c
```diff
@@ -544,10 +544,25 @@ static int tpm_start(uint32_t flags, TPMLIB_TPMVersion l_tpmversion,
...
 
     *res = TPM_FAIL;
 
+    /*
+     * swtpm_cuse only supports the dir:// backend. uri is untrusted
+     * (it comes from --tpmstate backend-uri=... on the command line);
+     * advancing past "dir://" without checking it's actually present
+     * and long enough leaves tpmdir pointing past the end of the
+     * allocation for any shorter/mismatched value.
+     */
```
stefanberger: The comment is a bit too long for this bug. Can you remove it?

Johnsonajibi: Done, condensed to a single line. Pushed as 72018ac.

stefanberger:
Thanks!

